PR #3, opened 2018-07-11
Added badges

jkuruvilla:
Added crates.io, docs.rs and license badges.


njaremko:
Saw this but forgot about it. Thanks :)

PR #5, opened 2018-07-22
Documentation test fix for Travis CI

jkuruvilla:
- Updated the example in README to match the one from the documentation.
- Updated the doc test in lib.rs to have a successful build in Travis CI
- Example of a successful build: https://travis-ci.org/jkuruvilla/iota-lib-rs/jobs/406766203

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,7 @@ Here's an example of how to send a transaction: (Note that we're using the addre
 extern crate iota_lib_rs;
 
 use iota_lib_rs::iota_api;
+use iota_lib_rs::iota_api::SendTransferOptions;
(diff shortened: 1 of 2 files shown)


jkuruvilla:
Okay both checks seem to fail now, while the Travis CI from my local repo was able to pass. Not sure why same test at two different repo is giving two different results.

njaremko:
The issue is that the end of that test makes a transaction on the tangle using local PoW, which might be too much for CI, so I might change to remote PoW.

Also, the majority of the tangle being out of sync right now is not helping.

jkuruvilla:
Ah alright, I shall close this then. Seems like remote PoW would be the way to go to sort this out.

PR #7, opened 2019-06-29
More documentation for iota-constants

wusyong:
Add a little bit more documentation for iota-constants. There are a few are not used right now, I'm not sure it should be cleaned up or not.

### DIFF
--- a/iota-constants/lib.rs
+++ b/iota-constants/lib.rs
@@ -7,6 +7,7 @@ pub const TRYTE_ALPHABET: [char; 27] = [
     'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z',
 ];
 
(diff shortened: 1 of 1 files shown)


iota-constants/lib.rs
```diff
@@ -132,6 +155,7 @@ pub const PRIVATE_KEY_REUSE_ERROR: &str = "Private key reuse detect!";
 /// SEND_TO_INPUTS_ERROR
 pub const SEND_TO_INPUTS_ERROR: &str = "Send to inputs!";
 
+/// SecurityLevel of IOTA signature. 2 is used mostly.
```
njaremko: The rest of these comments are fine, but this one needs to be more in-depth. Say the consequences of the different security levels.

njaremko (approved):

PR #17, opened 2019-09-02
client: merge iota_api into iota_client

### LINKED ISSUE #13: client: Move extended API to iota-client
At the moment higher level APIs are under `src` directory. It would be better they are all under same crate with basic IRI APIs.

wusyong:
Resolve #13


oopsmonk:
Can you separate responses API into core and extended? 
For instance, `iota-client/core/add_neighbors.rs` contains both response and request implementation.

oopsmonk (approved):

PR #18, opened 2019-09-02
client: fix bundle index

### LINKED ISSUE #16: Transfer doesn't create a bundle with multiple transactions
If I send a transaction with a long message that doesn't fit into a single signatureMessageFragement, a second transaction is created with the same bundle hash, but at the same index 0 instead of 1:
https://thetangle.org/bundle/CGGYHCWAABKSHHBIQYIIOOHKKKQSFBXRJOEKEXGNVVHVLEHIIUKPBBSHFNSHKCPLEAKGZEVKZHHHPCGPZ

wusyong:
Fix #16

### DIFF
--- a/iota-model/bundle.rs
+++ b/iota-model/bundle.rs
@@ -101,6 +101,18 @@ impl Bundle {
         }
     }
 
+    /// Reset bundle indexes
+    pub fn reset_indexes(&mut self) {
+        let mut current_index = 0;
+        let last_index = self.0.len() - 1;
+
+        for tx in &mut self.0 {
+            tx.current_index = current_index;
+            tx.last_index = last_index;
+            current_index += 1;
+        }
+    }
+
     /// Finalizes the bundle
     pub fn finalize(&mut self) -> Result<()> {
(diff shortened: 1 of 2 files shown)


iota-model/bundle.rs
```diff
@@ -101,6 +101,20 @@ impl Bundle {
         }
     }
 
+    /// Reset bundle indexes
+    pub fn reset_indexes(&mut self) -> Result<()> {
```
njaremko: Nothing in here returns an error, so why are you returning a result?

njaremko:
You shouldn't amend commits and force push in a PR, you mess with the historical record. You can choose to squash when the PR is merged

wusyong:
I think it's just preference. Personally if the change is trivial, I prefer to make commit log identical to the subject.

tsvisabo (approved):

oopsmonk:
> I think it's just preference. Personally if the change is trivial, I prefer to make commit log identical to the subject.

I agree with @njaremko, you should use `squash and merge`.

PR #25, opened 2019-10-01
client: add core API test cases

### LINKED ISSUE #22: client: Core API test cases.
- [ ] add_neighbors
- [ ] attach_to_tangle
- [ ] broadcast_transactions
- [ ] check_consistency
- [ ] find_transactions
- [ ] get_balances
- [ ] get_inclution_states
- [ ] get_neighbors
- [ ] get_node_info
- [ ] get_tips
- [ ] get_transactions_to_approve
- [ ] get_trytes
- [ ] remove_neighbors
- [ ] store_transactions
- [ ] were_address_spent_from

wusyong:
Resolve #22


iota-client/tests/core_api_test.rs
```diff
@@ -0,0 +1,345 @@
+use iota_client::options::*;
+
+mod common;
+use crate::common::*;
+
+#[test]
+fn test_add_neighbors_empty() {
+    let mut client = client_init();
+    let res = client.add_neighbors(&vec!["".into()]).unwrap();
```
oopsmonk, line 9: should not accept an empty url?

iota-client/tests/core_api_test.rs
```diff
@@ -0,0 +1,345 @@
...
+        assert_eq!(*neighbor, 0);
+    }
+}
+
+#[test]
+fn test_attach_to_tangle_empty() {
+    let mut client = client_init();
+
+    let opt = AttachOptions::default();
+    let _res = client.attach_to_tangle(opt).unwrap_err();
```
oopsmonk: should validate the return?

iota-client/tests/core_api_test.rs
```diff
@@ -0,0 +1,345 @@
...
+#[test]
+fn test_attach_to_tangle_empty_trunk() {
+    let mut client = client_init();
+
+    let opt = AttachOptions {
+        branch_transaction: TEST_BRANCH_HASH,
+        trytes: &[TEST_TX_HASH.into()],
+        ..AttachOptions::default()
+    };
+    let _res = client.attach_to_tangle(opt).unwrap_err();
```
oopsmonk: ditto

iota-client/tests/core_api_test.rs
```diff
@@ -0,0 +1,345 @@
...
+#[test]
+fn test_attach_to_tangle_empty_branch() {
+    let mut client = client_init();
+
+    let opt = AttachOptions {
+        trunk_transaction: TEST_TRUNK_HASH,
+        trytes: &[TEST_TX_HASH.into()],
+        ..AttachOptions::default()
+    };
+    let _res = client.attach_to_tangle(opt).unwrap_err();
```
oopsmonk: ditto

iota-client/tests/core_api_test.rs
```diff
@@ -0,0 +1,345 @@
...
+}
+
+#[test]
+fn test_broadcast_transactions_empty() {
+    let mut client = client_init();
+    let _res = client.broadcast_transactions(&["".into()]).unwrap_err();
```
oopsmonk: ditto

iota-client/tests/core_api_test.rs
```diff
@@ -0,0 +1,345 @@
...
+}
+
+#[test]
+fn test_check_consistency_empty() {
+    let mut client = client_init();
+    let _res = client.check_consistency(&["".into()]).unwrap_err();
```
oopsmonk: ditto

iota-client/tests/core_api_test.rs
```diff
@@ -0,0 +1,345 @@
...
+}
+
+#[test]
+fn test_get_transactions_to_approve_invalid_depth() {
+    let mut client = client_init();
+    let opt = GetTransactionsToApproveOptions {
+        depth: usize::max_value(),
+        ..GetTransactionsToApproveOptions::default()
+    };
+    let _res = client.get_transactions_to_approve(opt).unwrap_err();
```
oopsmonk: ditto

iota-client/tests/core_api_test.rs
```diff
@@ -0,0 +1,345 @@
...
+fn test_get_trytes() {
+    let mut client = client_init();
+    let res = client.get_trytes(&[TEST_BUNDLE_TX_1.into()]).unwrap();
+    assert!(res.error().is_none());
+}
+
+#[test]
+fn test_get_trytes_empty() {
+    let mut client = client_init();
+    let _res = client.get_trytes(&["".into()]).unwrap_err();
```
oopsmonk: ditto

iota-client/tests/core_api_test.rs
```diff
@@ -0,0 +1,345 @@
...
+
+    if let Some(neighbor) = res.removed_neighbors() {
+        assert_eq!(*neighbor, 0);
+    }
+}
+
+#[test]
+fn test_store_transactions_empty() {
+    let mut client = client_init();
+    let _res = client.store_transactions(&["".into()]).unwrap_err();
```
oopsmonk: ditto

iota-client/tests/core_api_test.rs
```diff
@@ -0,0 +1,345 @@
...
+#[test]
+fn test_get_trytes_empty() {
+    let mut client = client_init();
+    let _res = client.get_trytes(&["".into()]).unwrap_err();
+}
+
+#[test]
+fn test_remove_neighbors_empty() {
+    let mut client = client_init();
+    let res = client.remove_neighbors(&vec!["".into()]).unwrap();
```
oopsmonk, line 342: should not accept an empty url?

oopsmonk (changes requested):

iota-client/tests/core_api_test.rs
wusyong: IRI API doesn't block this and would return 0 neighbor. Or maybe we add a check before sending it?

iota-client/tests/core_api_test.rs
wusyong: I think `unwrap_err` should do the work. It means it returns `Err(_)` in `Result` before sending it. Otherwise, it will panic if it receives Ok(_).

oopsmonk: yes, but it doesn't tell what error is. that means this case can cover `test_attach_to_tangle_empty_trunk` and `test_attach_to_tangle_empty_branch` as well. 
what I think is the error string should contain `Provided trunk transaction is not valid` here.
The test plan does not just yield errors but also validates errors.

iota-client/tests/core_api_test.rs
oopsmonk: sending out empty request is not make any sense for me, it's up to you since IRI accepted it.

iota-client/tests/core_api_test.rs
wusyong: I see. I will also check the error type is the right one.

iota-client/tests/core_api_test.rs
wusyong: I think I'll just remain what it is until any request is needed.

iota-client/core/get_node_info.rs
```diff
@@ -30,16 +30,16 @@ pub struct GetNodeInfoResponse {
     app_version: String,
     /// Number of threads IRI is using
     #[serde(rename = "jreAvailableProcessors")]
-    jre_available_processors: usize,
+    jre_available_processors: u64,
```
oopsmonk: this maximum processor number is crazy. maybe u16?

oopsmonk (changes requested):

iota-client/core/get_node_info.rs
wusyong: sorry, I saw jre and just assigned it to u64 😂

oopsmonk (approved):

PR #31, opened 2019-11-27
update example

Thoralf-M:


wusyong (approved):
Thanks! Sorry I didn't review it right away, we are trying to proposed new crates in bee.

PR #32, opened 2019-11-28
check for kerl address in value transfers

Thoralf-M:


iota-client/extended/prepare_transfers.rs
```diff
@@ -87,6 +87,16 @@ impl<'a> Client<'a> {
                 );
             }
             transfer.address = iota_signing::checksum::remove_checksum(&transfer.address);
+            if transfer.value > 0 {
+                ensure!(
+                    if transfer.address.trits()[242] == 0 {
+                        true
+                    } else {
+                        false
+                    },
```
wusyong: You don't need to have these since `transfer.address.trits()[242] == 0` is already a expression. Could you also only convert last tryte to trit, so we don't have to convert whole address?

iota-client/extended/prepare_transfers.rs
```diff
@@ -235,7 +245,14 @@ impl<'a> Client<'a> {
             let to_subtract = 0 - this_balance;
             let timestamp = Utc::now().timestamp();
             let address = iota_signing::checksum::remove_checksum(&input.address);
-
+            ensure!(
+                if address.trits()[242] == 0 {
```
wusyong: Ditto

iota-client/extended/prepare_transfers.rs
```diff
@@ -247,6 +264,14 @@ impl<'a> Client<'a> {
             if this_balance >= total_transfer_value {
                 let remainder = this_balance - total_transfer_value;
                 if let Some(remainder_address) = &options.remainder_address {
+                    ensure!(
+                        if remainder_address.trits()[242] == 0 {
```
wusyong: Ditto

wusyong (changes requested):

iota-client/extended/prepare_transfers.rs
Thoralf-M: I don't know how to remove this part, if I just remove it I get
```
mismatched types

expected bool, found ()

note: expected type `bool`
         found type `()
````


To only convert the last char is it okay to do it this way? 
```Rust
let last_char = &transfer.address[transfer.address.len()-1..].to_string();
if last_char.trits()[2] == 0

wusyong: Maybe you still have the `if` keyword, just leave expression like this should work:
`ensure!(transfer.address.trits()[242] == 0,"Invalid Kerl address.");`
On second thought about last tryte, I think let's just remain your original way since `String` is not able to index. It's more about the design of type address. I will look into this when common crate rewrite.

Thoralf-M: Thanks!

wusyong (approved):

PR #46, opened 2020-04-29
Redesign workspace structure

### LINKED ISSUE #26: Update README
We should sync README like libraries in other languages to let users easier to start with.

### LINKED ISSUE #45: Re-export bee crates
We should re-export the crates users need to handle API. Add examples to show how to use them

wusyong:
This PR adjusts whole infrastructure with `iota-sdk` as main crate with the library named as `iota` to let user easier to import. Some common crates from bee are re-exported.

I also update the README and CHANGELOG in prepare for alpha release.

Resolve #45 
Resolve #26


iota-client/tests/api_test.rs
```diff
@@ -6,7 +6,7 @@ use bee_signing::*;
 use bee_ternary::*;
 use iota_client::response::*;
 
-#[tokio::test]
+#[smol_potat::test]
```
bingyanglin, line 9: Could you simply explain why use `smol_potat` instead of `tokio`?

bingyanglin (commented):
LGTM! Would you simply explain why we use `smol_potat` instread of `tokio`?

iota-client/tests/api_test.rs
wusyong: It's new runtime lately which is compatible to all other runtime and framework. But instead of create a whole runtime by default, it create only an executor which definitely faster when we testing.
Btw, this is dev-dependency which users are not going to use. They can choose the one they fond of.

Cargo.toml
```diff
@@ -1,58 +1,8 @@
...
-    #"iota-pow",
-    #"iota-crypto",
-    #"iota-conversion",
-    #"iota-model",
-    #"iota-utils",
-    #"iota-constants",
-    #"iota-validation",
-    #"iota-signing",
-    "iota-client"
+    "iota-sdk",
```
oopsmonk: usually, SDK binds to hardware or relates to some hardware. maybe iota-components? or do we really need it? how about move iota-client to root dir?

examples/send-transfers.rs
```diff
@@ -0,0 +1,63 @@
...
+    });
+
+    // Create a client instance
+    let res = iota::Client::new("https://nodes.comnet.thetangle.org")
+        // Call send_transfers api
+        .send_transfers()
+        // Input your seed
+        .seed(
+            &IotaSeed::<Kerl>::from_buf(
+                TryteBuf::try_from_str("RVORZ9SIIP9RCYMREUIXXVPQIPHVCNPQ9HZWYKFWYWZRE9JQKG9REPKIASHUUECPSQO9JT9XNMVKWYGVA")
```
oopsmonk, line 46: the seed is the same as address might confuse users, adds a comment or use a dummy seed?

bingyanglin (approved):

Cargo.toml
wusyong: Since this is a workspace, I would like to make every crates be member of it. I would like to rename with `iota` but that name is already taken from crates.io . I also so want the name be easy to identify as main crate so sdk is the closest I came up with.

examples/send-transfers.rs
wusyong: Let me add more comment about this then.

Cargo.toml
wusyong: How about iota-core?

oopsmonk: sounds good

README.md
```diff
@@ -37,64 +37,47 @@ Using the library is fairly easy, just add it as dependancy in `Cargo.toml`:
...
+  let mut iota = iota::Client::new("https://nodes.comnet.thetangle.org");
   println!("{:#?}", iota.get_node_info().unwrap());
 }
 ```
 
 
 ## API reference
 
-For details on all available API methods, see the [documentation](https://docs.rs/iota-lib-rs).
+You can read the [API reference] here, or generate them on your own.
```
oopsmonk: missing link on `API reference`.

iota-core/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+name = "iota-core"
+version = "0.1.0-alpha.1"
+authors = ["YuWeiWu <wusyong9104@gmail.com>"]
+edition = "2018"
+
+[lib]
+name = "iota"
+
+[dependencies]
+bee-bundle = { git = "https://github.com/Alex6323/bee-p" }
```
oopsmonk, line 11: since iota-client depends on bee, does iota-core still need to add bee in dependency? same as the iota-conversion package.

wusyong: I want to give a horizontal view in core crate. Make it more like a workspace structure.

oopsmonk (approved):

Issue #44, opened 2020-04-28
[Task]: API Responses need to be validated in test cases. 


oopsmonk:
The API responses are not validated in test cases. Tests should make sure responses are in the expectation.

wusyong:
There will be various responses on different timing/network. That's why I unwrap most of them atm. It won't panic if they do receive responses from the node.  It's difficult to cover all the situations imho.

oopsmonk:
it doesn't mean to cover all responses from the node but expected responses. for instance, the milestone should not be an empty hash in the getNodeInfo response from a healthy node. some fields have reasonable value and predictable in a response. you can check out what are reasonable responses in https://github.com/iotaledger/iota.c/tree/master/cclient/api/tests

thibault-martinez:
Syntax is already validated by `bee-rest-types`.
The rest will be superseded by https://github.com/iotaledger/iota.rs/issues/1199.

PR #51, opened 2020-05-04
Add message transfers support

### LINKED ISSUE #48: Support message transfer
Deliver messages is not available for now. We should bring it back with message can be more than one transaction long.

wusyong:
Now we can send message to tangle. And the message can be longer than one tx's length.
This PR also provides a example to send multiple lorem ipsum paragraphs.
https://comnet.thetangle.org/transaction/TKDNNYZVICPGIIMENYYWJQHYXQGOZYRGWWIMWZIWCKSQHXOYMRUPABSKVZZMRQGLKLRZXHPQKSOQJB999
Close #48

### DIFF
--- a/examples/message-transfers.rs
+++ b/examples/message-transfers.rs
@@ -1,0 +1,75 @@
+//! Send a transfer to an address.
+//!
+//! Run with:
+//!
+//! ```
+//! cargo run --example message-transfers
+//! ```
+use anyhow::Result;
+use iota::{
+    bundle::{Address, TransactionField},
+    client::Transfer,
+    crypto::Kerl,
+    signing::{IotaSeed, Seed},
+    ternary::{T1B1Buf, TryteBuf},
+};
+use iota_conversion::Trinary;
+
+#[smol_potat::main]
+async fn main() -> Result<()> {
+    // Prepare a vector of transfers
+    let mut transfers = Vec::new();
+
+    // Push the transfer to vector.
+    transfers.push(Transfer {
+        // Address is 81 trytes.
+        address: Address::from_inner_unchecked(
+            TryteBuf::try_from_str(
+                "RVORZ9SIIP9RCYMREUIXXVPQIPHVCNPQ9HZWYKFWYWZRE9JQKG9REPKIASHUUECPSQO9JT9XNMVKWYGVA",
+            )
+            .unwrap()
+            .as_trits()
+            .encode(),
+        ),
+        // We are using a zero balance seed so we make a zero value transfer here
+        value: 0,
+        message: Some(String::from("
+
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla dapibus eros at tincidunt fringilla. Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere cubilia curae; Sed mattis orci felis, vel pulvinar est sodales non. Vestibulum placerat luctus dapibus. Quisque finibus lacus a mauris fermentum eleifend. Maecenas pharetra justo nunc, in egestas magna suscipit at. Vivamus dictum at est pharetra porttitor. Integer ac neque elementum, feugiat lorem a, suscipit nisi. Etiam vel vehicula ligula. In bibendum aliquet vulputate. Pellentesque tincidunt elementum convallis.
+        
+        Maecenas auctor erat quis gravida pellentesque. Nam posuere gravida nisl. Ut tincidunt vitae neque placerat varius. Aliquam eget ultrices sem. Sed iaculis convallis magna at porttitor. Ut blandit ut ipsum id pellentesque. Vestibulum a egestas nisi, at auctor arcu. Nulla bibendum ut magna nec elementum. In sed ipsum nec nunc tincidunt porttitor. Sed sed aliquam enim. Sed ut elementum sapien, sit amet porta magna. Donec eget nulla nec sem sodales mollis at vel magna. In rhoncus ornare lectus a aliquet.
+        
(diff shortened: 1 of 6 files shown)


examples/message-transfers.rs
```diff
@@ -0,0 +1,75 @@
...
+        // Input the transfers
+        .transfers(transfers)
+        // We are sending to comnet, so mwm should be 10. It's 14 by default if you don't call this.
+        .min_weight_magnitude(10)
+        // Sending to the node and receive the response
+        .send()
+        .await?;
+
+    // The response of send_transfers is vector of Transaction type. We choose the first one and see what is its bundle hash
+    println!("{:?}", res[0].bundle().to_inner().as_i8_slice().trytes());
```
bingyanglin, line 76: Can the bundle trytes be known beforehand? If so it's better to use assertion for the testing purpose.

examples/send-transfers.rs
```diff
@@ -58,7 +59,7 @@ async fn main() -> Result<()> {
         .await?;
 
     // The response of send_transfers is vector of Transaction type. We choose the first one and see what is its bundle hash
-    println!("{:?}", res[0].bundle());
+    println!("{:?}", res[0].bundle().to_inner().as_i8_slice().trytes());
```
bingyanglin, line 63: S/A

examples/message-transfers.rs
```diff
@@ -0,0 +1,75 @@
+//! Send a transfer to an address.
+//!
+//! Run with:
+//!
+//! ```
+//! cargo run --example message-transfers
```
oopsmonk, line 6: example target is needed in Cargo.toml.

examples/message-transfers.rs
```diff
@@ -0,0 +1,75 @@
...
+        Aenean ultrices porttitor enim. Vestibulum vel ante at neque laoreet sollicitudin ut nec purus. Donec tortor lectus, egestas non gravida nec, consequat eu orci. Sed id vehicula eros. Aliquam leo odio, finibus nec placerat sit amet, dignissim in est. Suspendisse in neque auctor, bibendum nunc non, luctus dolor. Nulla convallis felis ac libero egestas, at egestas nunc consequat. Suspendisse tellus massa, consequat non felis facilisis, maximus convallis tortor. Fusce porta nunc est, et facilisis dolor imperdiet vitae. Vivamus ullamcorper erat et dignissim mattis. Nulla maximus, nisi vel lobortis gravida, arcu neque blandit ante, at sodales risus enim a metus. Nulla quis finibus mauris. ")),
+    });
+
+    // Create a client instance
+    let res = iota::Client::new("https://nodes.comnet.thetangle.org")
+        // Call send_transfers api
+        .send_transfers()
+        // Below is just a dummy seed which just serves as an example.
+        // If you want to replace your own. It probably should be a seed with balance on comnet/devnet.
+        .seed(
```
oopsmonk, line 59: the seed is useless when sending zero-transaction, therefore, no reason to provide seed here.

examples/message-transfers.rs
```diff
@@ -0,0 +1,75 @@
...
+                    .as_trits()
+                    .encode::<T1B1Buf>(),
+            ).unwrap(),
+        )
+        // Input the transfers
+        .transfers(transfers)
+        // We are sending to comnet, so mwm should be 10. It's 14 by default if you don't call this.
+        .min_weight_magnitude(10)
+        // Sending to the node and receive the response
+        .send()
```
oopsmonk, line 72: how can I set the tag in the transaction?

iota-client/tests/api_test.rs
```diff
@@ -378,7 +378,7 @@ async fn test_prepare_transfers_no_value() {
         transfers.push(Transfer {
             address: Address::zeros(),
             value: 0,
-            message: "".to_string(),
+            message: None,
```
oopsmonk, line 381: the tag field is needed in transfer implementation.

oopsmonk (changes requested):

examples/message-transfers.rs
wusyong: But this is not a test. What you mention should be done in api_test.rs which related to #44

examples/message-transfers.rs
wusyong: cargo can auto detect examples folder. So this command should work by default.

examples/message-transfers.rs
wusyong: It's not supported atm. And I think it's better left as is for now since we have been talking tag is not reliable for a while.

examples/message-transfers.rs
wusyong: Providing seed is still mandatory for now. I think this should file an issue instead.

examples/message-transfers.rs
oopsmonk: I believe you mixed up `examples package` and `examples folder` here. you created an examples package in this workspace if you want Cargo to find your examples you should locate them in `iota.rs/examples/examples` folder. have you tried to run `cargo run --example message-transfers` via CLI? 
if you can be mixed up with it then users too, my suggestion is to rename this examples package to iota-client-examples or something else as you prefer.

examples/message-transfers.rs
oopsmonk: tag is not reliable on what? afaik, tag is an optional field and it's user-defined.

examples/message-transfers.rs
bingyanglin: Got it. Then agree to remain the same.

iota-client/tests/api_test.rs
bingyanglin: Also can we set the timestamp here or not needed?

wusyong: No, timestamp should be handled by method itself. Regarding tag with above comment, I'll push it in a moment then.

examples/message-transfers.rs
wusyong: I think I just forgot to add this example to the cargo.toml. I would like to leave this as is. It should work on outside workspace and under examples folder.

oopsmonk: alright, that was what I mean example target in Cargo.toml at first comment.

oopsmonk (approved):

examples/message-transfers.rs
oopsmonk: #52

PR #54, opened 2020-05-06
Add generate address examples

wusyong:


examples/get_new_address.rs
```diff
@@ -0,0 +1,43 @@
+//! Get a valid new address as transfer address.
```
oopsmonk:
```suggestion
//! Get an unused address from a connected node.
```

examples/get_unchecked_address.rs
```diff
@@ -0,0 +1,48 @@
+//! Get an unchecked from seed. This is just for developing purpose in case you want to see certain address.
```
oopsmonk:
```suggestion
//! Get an address from seed and index that the address could be an used address on the Tangle. This is just for developing purpose in case you want to see a certain address.
```

examples/get_unchecked_address.rs
```diff
@@ -0,0 +1,48 @@
...
+use iota::crypto::Kerl;
+use iota::signing::{
+    IotaSeed, PrivateKey, PrivateKeyGenerator, PublicKey, Seed, WotsPrivateKeyGeneratorBuilder,
+    WotsSecurityLevel,
+};
+use iota::ternary::{T1B1Buf, TryteBuf};
+use iota_conversion::Trinary;
+
+#[smol_potat::main]
+async fn main() -> Result<()> {
```
oopsmonk: async runtime is not needed in this case.
```suggestion
fn main() {
```

oopsmonk (changes requested):

oopsmonk (approved):

PR #60, opened 2020-05-12
Remove lifetime requirement on Client structure

wusyong:
This will make users no need to worry lifetime anymore and help wasm bindings easier.


bingyanglin (approved):
I like these changes.

PR #69, opened 2020-05-19
fix(wasm): Update methods to reflect refactoring in iota-client

rajivshah3:
Fixes WASM build for following methods that were affected by `Client` refactoring:

- `getNodeInfo`
- `getNewAddress`
- `attachToTangle`
- `broadcastBundle`
- `checkConsistency`
- `sendTransfers`

`addNeighbors` is partially fixed, I'm not entirely sure how to change it so that it passes a `&[&str]` instead of a `Vec<String>`

### DIFF
--- a/bindings/wasm/src/lib.rs
+++ b/bindings/wasm/src/lib.rs
@@ -87,7 +87,7 @@ impl Client {
 
     #[wasm_bindgen(js_name = "getNodeInfo")]
     pub async fn get_node_info(self) -> Result<JsValue, JsValue> {
-        let node_info = self.client.get_node_info()
+        let node_info = iota::client::client::Client::get_node_info()
             .await
             .map_err(|e| JsValue::from(e.to_string()))?;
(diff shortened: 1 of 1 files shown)


bindings/wasm/src/lib.rs
```diff
@@ -87,7 +87,7 @@ impl Client {
 
     #[wasm_bindgen(js_name = "getNodeInfo")]
     pub async fn get_node_info(self) -> Result<JsValue, JsValue> {
-        let node_info = self.client.get_node_info()
+        let node_info = iota::client::client::Client::get_node_info()
```
wusyong, line 90: I think `iota-core` already re-export the `Client`. You don't need to add the path that long.
```suggestion
        let node_info = iota::Client::get_node_info()
```

lucasfernog:
I've already implemented this fix, I didn't push it yet because I'm adding more APIs to the WASM binding.

PR #70, opened 2020-05-19
feat(bindings) extend API, update to new client interface, add CI test

lucasfernog:


bindings/wasm/src/lib.rs
```diff
@@ -66,208 +61,564 @@ fn js_error<T: std::fmt::Debug>(e: T) -> JsValue {
...
+        .map_err(js_error)?;
+    Ok(())
+}
 
-        let mut builder = self.client.get_new_address();
-        builder = builder.seed(&encoded_seed);
+#[wasm_bindgen(js_name = "addNeighbors")]
+pub async fn add_neighbors(uris: JsValue) -> Result<JsValue, JsValue> {
+    let uris: Vec<String> = uris.into_serde().map_err(js_error)?;
+    let added_neighbords = iota::Client::add_neighbors(uris)
```
rajivshah3:
```suggestion
    let added_neighbors = iota::Client::add_neighbors(uris)
```

bindings/wasm/src/lib.rs
```diff
@@ -66,208 +61,564 @@ fn js_error<T: std::fmt::Debug>(e: T) -> JsValue {
...
+        .await
+        .map_err(js_error)?;
 
-        if let Some(index) = index {
-            builder = builder.index(index);
-        }
-        if let Some(security) = security {
-            builder = builder.security(security);
-        }
+    let res = response_to_js_value(added_neighbords)?;
```
rajivshah3:
```suggestion
    let res = response_to_js_value(added_neighbors)?;
```

bindings/wasm/src/lib.rs
lucasfernog: You're the official typo finder :D thanks

iota-client/client.rs
```diff
@@ -72,9 +72,9 @@ impl Client {
     /// * [`uris`] - Slices of neighbor URIs(`&str`) to add
     ///
     /// [`uris`]: ../core/struct.AddNeighborsBuilder.html#method.uris
-    pub async fn add_neighbors(uris: &[&str]) -> Result<AddNeighborsResponse> {
-        for uri in uris {
-            match Url::parse(uri)?.scheme() {
+    pub async fn add_neighbors(uris: Vec<String>) -> Result<AddNeighborsResponse> {
```
wusyong, line 75: Is it has to be Vector? I was trying to avoid allocation as possible.

bindings/wasm/src/lib.rs
```diff
@@ -66,208 +61,564 @@ fn js_error<T: std::fmt::Debug>(e: T) -> JsValue {
...
+}
 
-    #[wasm_bindgen(js_name = "getNodeInfo")]
-    pub async fn get_node_info(self) -> Result<JsValue, JsValue> {
-        let node_info = self.client.get_node_info()
-            .await
-            .map_err(|e| JsValue::from(e.to_string()))?;
-        let res = response_to_js_value(node_info)?;
-        Ok(res)
+fn create_address(address: String) -> Result<Address, JsValue> {
```
wusyong, line 88: These functions are because those types are lack of serialize trait right? Could you add a TODO here to remind us to implement this later?

iota-client/response.rs
```diff
@@ -2,6 +2,59 @@
 use anyhow::Result;
 use iota_bundle_preview::{Address, Hash, Tag, Transaction, TransactionField};
 use iota_ternary_preview::TryteBuf;
+use serde::ser::{Serialize, Serializer, SerializeSeq, SerializeStruct};
+
+#[derive(Serialize)]
+pub struct TransactionDef {
```
wusyong, line 10: I enforce a lint to write documentation. This one is needed for example.

lucasfernog: yeah I saw that but forgot to add it.

iota-client/client.rs
lucasfernog: I tried to convert a JsValue to a `&[&str]` but couldn't do it, maybe you could give it a try, I'm out of ideas there.

bindings/wasm/src/lib.rs
lucasfernog: actually these functions are here to convert a javascript value (e.g. a string) to its Rust type (e.g. the Address).

the code that I had to write only because some types doesn't implement Serialize are the `impl Serialize for T` on client/response.rs

iota-client/client.rs
wusyong: Hmm, you are right. I was thinking `Box<[JsValue]>`, but it will do the allocation no matter what.

wusyong (approved):

PR #71, opened 2020-05-20
feat(bindings) extend C APIs & refactor to be more idiomatic in C

wusyong:


bindings/c/src/models.rs
```diff
@@ -0,0 +1,82 @@
...
+    Box::into_raw(Box::new(CSeed(IotaSeed::new())))
+}
+
+#[no_mangle]
+pub extern "C" fn iota_seed_free(ptr: *mut CSeed) {
+    if ptr.is_null() {
+        return;
+    }
+    unsafe {
+        Box::from_raw(ptr);
```
rajivshah3: Shouldn't we use [`zeroize`](https://docs.rs/zeroize/1.1.0/zeroize/trait.Zeroize.html#tymethod.zeroize) to zero the memory before freeing the pointer?

wusyong: Thanks for the advise! I think we can even do it ourselves since it just one volatile and one atomic operation.

PR #72, opened 2020-05-21
support more ascii chars and uneven trytes length

Thoralf-M:


Thoralf-M:
If we allow an uneven Trytes length we can just get the the Trytes from the signatureMessageFragment field and convert it to a message without the need of reducing the size to an even length before

wusyong (approved):
LGTM, thanks!

PR #74, opened 2020-05-22
chore: Fix some clippy warnings

rajivshah3:


iota-conversion/trinary.rs
```diff
@@ -127,7 +127,7 @@ impl Trinary for &str {
         trits_with_length(&self.trits(), length)
     }
     fn trytes(&self) -> Result<Trytes> {
-        Ok(self.to_string())
+        Ok((*self).to_string())
```
rajivshah3, line 130: > `&str` implements `ToString` through a slower blanket impl, but `str` has a fast specialization of `ToString`

https://rust-lang.github.io/rust-clippy/master/index.html#inefficient_to_string

iota-client/client.rs
```diff
@@ -60,7 +60,7 @@ impl Client {
     pub(crate) fn get_node() -> Result<Url> {
         Ok(
             unsafe { Client::get().uri.load(Ordering::Relaxed).as_ref() }
-                .ok_or(anyhow!("Fail to get node url"))?
+                .ok_or_else(|| anyhow!("Fail to get node url"))?
```
rajivshah3, line 63: > The function will always be called and potentially allocate an object acting as the default.

https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call

wusyong (approved):
Thanks, I removed CI for fromat&lint because we need to merge some bindings crate asap. I will bring it back next week.

PR #79, opened 2020-06-02
feat(wasm) prepare package for alpha publish

lucasfernog:


bindings/wasm/package.json
```diff
@@ -14,8 +17,14 @@
     "prepublishOnly": "yarn build",
     "test": "mocha"
   },
-  "author": "Lucas Nogueira <lucas.nogueira@iota.org>",
-  "license": "ISC",
+  "contributors": [
+    "Lucas Nogueira <lucas.nogueira@iota.org>"
+  ],
+  "license": "Apache-2.0",
```
wusyong: Why is it not MIT here?

wusyong (commented):
We probably should mention license in Cargo.toml too.

bindings/wasm/package.json
nothingismagick: My understanding is that Apache-2.0 is the license for all IOTA projects moving forward, but yes, we should use also use MIT here for historical reasons... (original license is MIT) 

so: MIT or Apache 2 (just like the rust ecosystem) ?

wusyong: yeah, I think it would be great to be MIT or Apache2 like what we usually do on general crates.

Issue #89, opened 2020-06-24
trytes_converter only handles ASCII


mayjs:
Is there a reason why we can't just convert arbitrary binary data to trytes? The whole trytes_converter module seems overly complex to me, why do we have to use manual lookup tables from ascii to numerical values instead of just using `str.bytes` and converting? The latter change would also allow us to convert utf-8 to trytes.

wusyong:
It's more a historical reason that it's is ASCII only. I'm actually preparing a module to support more encodings (with utf8 as default).

cvarley100:
We will not be making additional changes to the existing mainnet implementation. The current focus is Chrysalis.

PR #90, opened 2020-06-24
Refactored trytes_converter to work for utf-8 strings and binary data

mayjs:
I've implemented a proof of concept for my proposal in issue #89. I also got rid of the `failure` errors, since `failure` is deprecated and I think it's better to use proper error types for library code anyways.
We could also use something like the `thiserror` crate for automatic error implementations without manual `Display` and `From` implementations.


wusyong (approved):
Thanks, LGTM! We are actually migrating to more modular approach to convert strings. So this is really helpful.

PR #94, opened 2020-07-02
fix(wasm) expose f64 instead of u64 so it works on Safari/iOS

lucasfernog:
Apple doesn't support the [BigUint64Array API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigUint64Array#Browser_compatibility) so I have to move to f64 to make it work on Safari/iOS.


wusyong (approved):
LGTM. I actually got some feedback that there are some feedback about this earlier today.

PR #99, opened 2020-07-13
feat(wasm) expose `findTransactions` and `getTrytes`

lucasfernog:


wusyong (approved):
LGTM. But I would treat this as temporary measure.

PR #101, opened 2020-07-15
Fixes typo in the exports of the WASM web bindings

kuhlmannmarkus:


nothingismagick:
not sure i agree with this semantically, because camelCase is used to demarcate different words. In my book `broadcast` is always one word and i would never intuit that it should be written `broadCast`. nevertheless, this should be the same everywhere so your concern is definitely valid.

kuhlmannmarkus:
I think you are right @nothingismagick

wusyong:
Well... `broadcast` is a word in my understanding too. Besides, I followed the naming rule in [iota.js](https://github.com/iotaledger/iota.js/#api-reference).

wusyong:
@kuhlmannmarkus thanks for the PR. I think it's the function name has typo instead of export. But I think it's good to merge as a hot fix.

PR #119, opened 2020-08-06
fix(crate) compilation for android targets

lucasfernog:
The `openssl` crate (a dependency of reqwest) doesn't cross-compile well, so this PR disables it. @wusyong is this ok to do? maybe we should pick another TLS solution from reqwest? https://github.com/seanmonstar/reqwest/blob/83fa93ccafe1ec5044f2edd118006edf4faa28b2/Cargo.toml#L25

Also tried to disable it only when compiling to Android (using the `[target.cfg().dependencies]` option) but somehow it didn't work.

### DIFF
--- a/iota-client/Cargo.toml
+++ b/iota-client/Cargo.toml
@@ -17,7 +17,7 @@ bee-transaction = { git = "https://github.com/Alex6323/bee-p.git" }
 bee-ternary = { version = "0.3.1-alpha", git = "https://github.com/iotaledger/bee.git", features = ["serde1"] }
 bee-signing = { git = "https://github.com/iotaledger/bee.git" }
 bee-crypto = { git = "https://github.com/iotaledger/bee.git" }
-reqwest = { version = "0.10.6", features = ["json"] }
+reqwest = { version = "0.10.6", features = ["json"], default-features = false }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 chrono = "0.4.12"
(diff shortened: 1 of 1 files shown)


iota-client/Cargo.toml
```diff
@@ -17,7 +17,7 @@ bee-transaction = { git = "https://github.com/Alex6323/bee-p.git" }
 bee-ternary = { version = "0.3.1-alpha", git = "https://github.com/iotaledger/bee.git", features = ["serde1"] }
 bee-signing = { git = "https://github.com/iotaledger/bee.git" }
 bee-crypto = { git = "https://github.com/iotaledger/bee.git" }
-reqwest = { version = "0.10.6", features = ["json"] }
+reqwest = { version = "0.10.6", features = ["json"], default-features = false }
```
wusyong: It still have to choose one of tls library. The most sane choice I think is rustls. But it might bloat the size a bit.

```suggestion
reqwest = { version = "0.10.6", features = ["json", "rustls-tls"], default-features = false }
```

lucasfernog: yeah I see a jump from 5.5MB to 7.1MB when using rustls-tls, so we should use it only when compiling to Android (at least for now).

lucasfernog:
We should keep an eye on https://github.com/rust-lang/cargo/issues/1197#issuecomment-590385530

to properly define the reqwest dependency as this:

```
[target.'cfg(target_os = "android")'.dependencies]
reqwest = { version = "0.10.6", default-features = false, features = ["json", "rustls-tls"] }

[target.'cfg(not(target_os = "android"))'.dependencies]
reqwest = { version = "0.10.6", default-features = false, features = ["json", "default-tls"] }
```

lucasfernog:
Tested to compile for Android, works fine now. But I had to add my NDK/x86/bin to my PATH env because of the `ring` crate compilation.

wusyong (approved):

PR #143, opened 2020-10-20
add indexation example and get real values, update dependency

Thoralf-M:


wusyong (approved):
Ah I didn't aware there's already a PR for this. I'll merge this then. Thanks! (and also the formatting)

PR #145, opened 2020-10-26
Update specification to make API more friendly

wusyong:
Address several changes to make it more user friendly, so they don't need to know too many details when using high level APIs:

- All `BIP32Path` parameters are optional. Since address derivation rfc doesn't come out yet. I would like to remain as is until the certain method is defined.
- `indexation` parameters is splited into `indexation_key` and `data`.
- Move GET method of address, output, and message back to low level. Users probable just want to use find_* API most of the time.

The amount must still be non zero is because transaction payload must have an output and its amount must not be zero during [syntactical validation](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md#syntactical-validation).
And the last thing I haven't updated yet is `reattach()`. @fijter  Do you mean it should be moved to `send()` method as optional parameter?

### DIFF
--- a/specs/iota-rs-ENGINEERING-SPEC-0000.md
+++ b/specs/iota-rs-ENGINEERING-SPEC-0000.md
@@ -18,6 +18,7 @@ Specification of High Level Abstraction API
   * [`get_unspent_address`](#get_unspent_address)
   * [`get_balance`](#get_balance)
   * [`get_address_balances`](#get_address_balances)
+  * [`retry`](#retry)
 * [Full Node API](#Full-Node-API)
   * [`get_health`](#get_health)
   * [`get_info`](#get_info)
@@ -28,6 +29,8 @@ Specification of High Level Abstraction API
   * [`find_outputs`](#find_outputs)
   * [`find_addresses`](#find_addresses)
   * [`get_milestone`](#get_milestone)
+  * [`reattach`](#reattach)
+  * [`promote`](#promote)
 * [Objects](#Objects)
   * [Network]
   * [Hash]
@@ -222,6 +225,17 @@ Following are the steps for implementing this method:
     parameter;
 *   Return the list of Output which contains corresponding pairs of address and balance.
 
+## `retry()`
+
+Retries (promotes or reattaches) a message for provided message id. Message should only be retried only if they are valid and haven't been
+confirmed for a while. 
+
+### Parameters
+
+| Field | Required | Type | Definition |
+| - | - | - | - |
+| **message_id** | ✔ | [MessageId] | The identifier of message. |
+
 ### Returns:
 
 Newly reattached [Message](#Message).
@@ -230,11 +244,9 @@ Newly reattached [Message](#Message).
 
 Following are the steps for implementing this method: 
 
-* Only unconfirmed messages should be allowed to reattach. The method should validate the confirmation state of the provided messages. If a message hash of a confirmed message is provided, the method should error out;
-* The method should also validate if the message reattachment is necessary. This can be done by checking if the message falls below max depth. The criteria of checking whether the message has fallen below max depth is through time. If 11 minutes have passed since the timestamp of the most recent (reattachment), the message can be allowed to be reattached. See [this](https://github.com/iotaledger/trinity-wallet/blob/3fab4f671c97e805a2b0ade99b4abb8b508c2842/src/shared/libs/iota/transfers.js#L141) implementation for reference;
-* Get tips pair using [`get_tips()`](#get_tips-get-tips);
-* Perform proof-of-work;
-* Store messages on the tangle using [`post_messages()`](#post_messages-post-messages);
+* Only unconfirmed messages should be allowed to retry. The method should validate the confirmation state of the provided messages. If a message id of a confirmed message is provided, the method should error out;    
+* The method should also validate if a retry is necessary. This can be done by leveraging the `/messages/{messageId}/metadata` endpoint (already available through [get_message](#get_message)). See [this](https://github.com/iotaledger/trinity-wallet/blob/develop/src/shared/libs/iota/transfers.js#L105-L131) implementation for reference;
+* Use [reattach](#reattach) or [promote](#promote) accordingly.
 
 # Low level Node API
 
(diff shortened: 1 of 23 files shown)


specs/iota-rs-ENGINEERING-SPEC-0000.md
```diff
@@ -156,7 +157,7 @@ Following are the steps for implementing this method:
 * Repeat the above step till there's an unspent address found;
 * Return the address with corresponding index on the wallet chain;
 
-## `get_addresses()`
+## `find_addresses()`
```
fijter: This function can probably be lower level instead of higher level - On a higher level you probably don't care about this, you just want to send and receive coins/data.

specs/iota-rs-ENGINEERING-SPEC-0000.md
```diff
@@ -79,11 +79,12 @@ A generic send function for easily sending a value transaction message.
...
 | - | - | - | - |
 | **seed** | ✔ | [Seed] | The seed of the account we are going to spend. |
-| **path** | ✔ | [BIP32Path] | The wallet chain BIP32 path we want to search for. |
 | **address** | ✔ | [Address] | The address to send to. |
 | **value** | ✔ | std::num::NonZeroU64 | The amount of IOTA to send. It is type of NoneZero types, so it cannot be zero. |
+| **path** | ✘ | [BIP32Path] | The wallet chain BIP32 path we want to search for. |
 | **output** | ✘ | Output | Users can manually pick their own output instead of having node decide on which output should be used. |
-| **indexation** | ✘ | Indexation | An optional indexation payload with indexation key and data. Both fields can be optional too. |
+| **indexation_key** | ✘ | String | An optional indexation key of the indexation payload. |
+| **data** | ✘ | String | An optional indexation data of the indexation payload. |
```
fijter: Semantics but I think both can be bytes (a String is just UTF-8 encoded bytes ofc, data or key can be in a binary representation as well iirc.)

specs/iota-rs-ENGINEERING-SPEC-0000.md
```diff
@@ -253,63 +254,6 @@ Following are the steps for implementing this method:
 * Perform proof-of-work;
 * Store messages on the tangle using [`post_messages()`](#post_messages-post-messages);
 
-## `get_message()`
```
fijter, line 256: Should probably stay in the higher level since this is a common thing to do, retrieving a single message by message id.

fijter (commented):
In addition to the comments in the code some other remarks:

 - Network in the constructor is not required, but you need to either have a network or a node/nodes/node_pool_urls list, one of those is mandatory but it doesn't matter which one.
 - `send` without value should be possible, in that case you don't send a Message with SignedTransaction with Indexation Payload but a Message with just a Indexation Payload instead and the Data in the value field of the indexation.
 - Maybe name message_id `message_ids` in `find_messages` since we always expect a list of MessageId's there?
 - Reattach could be a function on the message object returned by get_message/find_messages/send instead of a seperate api function. Does not make sense to add this as a optional attribute to send, you want to be able to re-attach any message regardless of if you've sent it yourself and if you actually called the send function in code or not.
 - find_ouputs and maybe find_addresses can probably move to lower level libs since I don't see this as something someone would commonly do on the higher level.

specs/iota-rs-ENGINEERING-SPEC-0000.md
wusyong: Their type definition in RFC are string and bytes array respectively. I would change the data parameter to u8 bytes.

fijter (commented):
Almost there I guess, some final points:

 - `find_messages` back to Higher level API part please, that was okay there (just leave find_outputs in Full Node API part)
 - `get_message` should only be lookup by message_id, not index (since one index can have multiple messages, find_messages does allow index lookup, corrosponding to the API docs for this at /api/v1/messages/{messageId})
 - The type of `get_address_balances` `addresses` parameter is `[MessageId]`, should probably be [Address]

fijter (approved):
Nice, looks good to me at this point. Looking forward to any feedback from others utilizing this to see if we missed anything further.

fijter:
Ah one more thing I just noticed; `output` should be `outputs` for the `send` method since you can provide an array of multiple outputs to use, not just a single one.

cvarley100:
@fijter I have updated the `output` and `address` parameters to arrays to allow for multiple. I suggest we keep the name the same.

lucasfernog:
When can we get the `reattach` function back?

lucasfernog:
Also we need to do something about this `is_confirmed` function - the wallet library needs to know that but it's not part of the node API spec (?)

Thoralf-M:
Why is `get_message` in the General API list and not in the Full Node API list?
And `find_addresses` is in the Full Node API, but it doesn't require the API at all

laumair:
> When can we get the `reattach` function back?

@lucasfernog Please see https://github.com/iotaledger/iota.rs/issues/151

> Also we need to do something about this is_confirmed function - the wallet library needs to know that but it's not part of the node API spec (?)

I do not think there will be a separate `is_confirmed` function. I believe [get_message](https://github.com/iotaledger/iota.rs/blob/dev/specs/iota-rs-ENGINEERING-SPEC-0000.md#get_message) when passed with `metadata()` option should provide us with the confirmation details of a message. The expected response interface should be:

```
{
    /**
     * The message id.
     */
    messageId: string;
    /**
     * The message id of parent 1.
     */
    parent1MessageId: string;
    /**
     * The message id of parent 2.
     */
    parent2MessageId: string;
    /**
     * Is the message solid.
     */
    isSolid?: boolean;
    /**
     * Is the message referenced by a milestone.
     */
    referencedByMilestoneIndex?: number;
    /**
     * The ledger inclusion state.
     */
    ledgerInclusionState?: string;
    /**
     * Should the message be promoted.
     */
    shouldPromote?: boolean;
    /**
     * Should the message be reattached.
     */
    shouldReattach?: boolean;
}
```

laumair:
@fijter @cvarley100 I couldn't find any information about address checksums. I believe there should be support for creating checksums for addresses (maybe on the lower level).

specs/iota-rs-ENGINEERING-SPEC-0000.md
```diff
@@ -69,6 +69,34 @@ The data structure to initialize the instance of the Higher level client library
...
+* Unsolved Questions 
+  - We make the iota.rs to be a static library, so whenever a user uses an API, then the `Client` instance should be built again, so as the `synced_node_list`. In this way we don't have to maintain the nodes status in the background.
+  - We set a timeout for the `Client` instance. If the time is out, then the `Client` instance is needed to perform the sync process again, so as to get the latest `synced_node_list`.
+  - We make the iota.rs maintain the up-to-date synced nodes. In this way we need a mechanism to update the each node status in the iota.rs, like 1) periodically queries healthy/unhealthy nodes 2) subscribe events from nodes 3) any other efficient syncing mechanism.
+  
+## List of APIs Whose Results Based on `quorum_size` Nodes
+
+* Unsolved Question:
+  - We need to define the list of APIs (e.g., `get_balance()`, `get_address_balances`, etc.), whose returned values should depend on the `quorum_size` nodes.
+
```
bingyanglin: Listed the open questions

PR #154, opened 2020-11-13
feat(client) add MQTT subscribe/unsubscribe APIs

lucasfernog:
This PR introduces the integration with Hornet's MQTT interface. Here's an example of the usage:

```rust
use iota::{Client, Topic};

fn main() {
    let mut iota = Client::new() // Crate a client instance builder
        .node("http://0.0.0.0:14265") // Insert the node here
        .unwrap()
        .build()
        .unwrap();
    iota.subscriber()
        .topic(Topic::new("milestones/latest").unwrap())
        .topic(Topic::new("addresses/5d2684dfdbd52a2241e135c15f9c5ede2bca5599f810ddbc14c75948540345bf/outputs").unwrap())
        .subscribe(|event| println!("{:?}", event))
        .unwrap();
    loop {}
}

```

### DIFF
--- a/iota-client/src/node/mqtt.rs
+++ b/iota-client/src/node/mqtt.rs
@@ -1,12 +1,13 @@
 use crate::client::{Client, TopicEvent, TopicHandlerMap};
 use crate::Result;
-use regex::Regex;
-use rumqttc::{
-  Client as MqttClient, Connection as MqttConnection, Event, Incoming, MqttOptions, QoS,
+use paho_mqtt::{
+  Client as MqttClient, ConnectOptionsBuilder, CreateOptionsBuilder, MQTT_VERSION_3_1_1,
 };
+use regex::Regex;
 
 use std::convert::TryFrom;
 use std::sync::{Arc, Mutex};
+use std::time::Duration;
 
 macro_rules! lazy_static {
   ($init:expr => $type:ty) => {{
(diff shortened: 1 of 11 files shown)


iota-client/src/node/mqtt.rs
```diff
@@ -0,0 +1,207 @@
...
+        if let Some(handlers) = mqtt_topic_handlers_guard.get(&Topic(topic.clone())) {
+          let event = TopicEvent {
+            topic,
+            payload: message.payload_str().to_string(),
+          };
+          for handler in handlers {
+            handler(&event)
+          }
+        }
+      }
```
bingyanglin, line 118: Can you help check the while let loop will break when the all the topics are unsubscribed? (Like the mpsc `Receiver` will be dropped.) And it might be good to handle the corresponding Sender is disconnected.

iota-client/src/node/mqtt.rs
```diff
@@ -0,0 +1,207 @@
...
+        .as_ref()
+        .ok_or_else(|| crate::Error::MqttConnectionNotFound)
+    }
+  }
+}
+
+fn poll_mqtt(mqtt_topic_handlers: Arc<Mutex<TopicHandlerMap>>, client: &mut MqttClient) {
+  let receiver = client.start_consuming();
+  std::thread::spawn(move || {
+    while let Ok(message) = receiver.recv() {
```
bingyanglin, line 107: Can you help check the while let loop will break when the all the topics are unsubscribed? (Like the mpsc Receiver will be dropped.) And it might be good to handle the corresponding Sender is disconnected.

bingyanglin (changes requested):

iota-client/src/node/mqtt.rs
bingyanglin: Why I point this out is that I saw this [issue](https://github.com/eclipse/paho.mqtt.rust/issues/61).

lucasfernog: do you think we should close the connection when all the topics are unsubscribed? for performance reasons I think we could expose a method to do it, because for instance the wallet will unsubscribe to all topics and then subscribe again frequently (on every sync).

bingyanglin: If we will subscribe/unsubscribe frequently I think it's good to remain the connection. Just we might need to expose a method to let the user disconnect and let the Receiver channel to be dropped, so as to break the `while let` loop and let the spawned thread be shut down gracefully. Or do we already have a way to let the spawned thread gracefully shut down if the user does not want to receive mqtt messages?

lucasfernog: just pushed the changes for it, yeah I had to drop the mqtt client instance in order to break from the while loop and shut down the thread. thanks for pointing that out, nice catch :P

lucasfernog: I also added a `broker_options` field to the client builder with a flag for the automatic disconnection when the subscriptions are empty (defaults to true), is that ok @fijter ?

bingyanglin:
LGTM!

bingyanglin (approved):

Issue #158, opened 2020-11-16
Complete the need-to-modify/unimplemented APIs


bingyanglin:
The need-to-modify or unimplemented yet APIs follow

**General API**
- [x] `find_messages()`: not implemented yet, need a wrapper and use the already implemented low-level APIs.
- [x] `get_address_balances()`: Not implemented yet, need a wrapper and use the already implemented low-level APIs.
- [x] `retry()`: Not implemented yet.

**Full Node API**

- [x] `find_outputs()`: Not implemented yet, need a wrapper and use the already implemented low-level APIs.
- [x] `find_addresses()`: Need rename it from `get_addresses()`.
- [x] `reattach()`: Not implemented yet.
- [x] `promote()`: Not implemented yet.

All of the above are documented in the specification and should be implemented.

PR #161, opened 2020-11-16
Complete the need-to-modify/unimplemented APIs to align with the specification

### LINKED ISSUE #157: Align send() API with specification
Current `send()` implementation does not support the following parameters for a message with value transaction 
1. output (Users can manually pick their own output instead of having node decide on which output should be used.)
2. indexation_key (An optional indexation key of the indexation payload.)
3. data (An optional indexation data of the indexation payload.)

Also, current `output` parameter in the `send()` is actually the `address` parameter defined in the [spec](https://github.com/iotaledger/iota.rs/blob/dev/specs/iota-rs-ENGINEERING-SPEC-0000.md), which is needed to be rename as `address`.

bingyanglin:
Fix #157 #158

### DIFF
--- a/iota-client/src/client.rs
+++ b/iota-client/src/client.rs
@@ -175,6 +175,29 @@ impl Client {
             status => Err(Error::ResponseError(status)),
         }
     }
+    /// Find all outputs based on the requests criteria. This method will try to query multiple nodes if 
+    /// the request amount exceed individual node limit.
+    pub async fn find_outputs(&self, outputs: &[UTXOInput], addresses: &[Address]) -> Result<Vec<OutputMetadata>> {
+        let mut output_metadata = Vec::<OutputMetadata>::new();
+        
+        // Use `get_output` API to get the `OutputMetadata`
+        for output in outputs {
+            let meta_data = self.get_output(output).await?;
+            output_metadata.push(meta_data);
+        }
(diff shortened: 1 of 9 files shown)


iota-client/src/client.rs
```diff
@@ -175,6 +175,29 @@ impl Client {
...
+        for output in outputs {
+            let meta_data = self.get_output(output).await?;
+            output_metadata.push(meta_data);
+        }
+
+        // Use `get_address()` API to get the address outputs first, then use the `get_output()`
+        // to get the `OutputMetadata`
+        for address in addresses {
+            let address_outputs = self.get_address().outputs(&address).await?;
+            for output in address_outputs.iter()  {
```
lucasfernog: should we check if the `outputs` parameter contains the output here to prevent duplication?

bingyanglin: Good point. Will use a `HashSet` to collect all the queried outputs first. Thank you for pointing this out.

lucasfernog (approved):

lucasfernog:
@bingyanglin can you comment out the `with_network_id` call? it was temporarily removed from bee.

lucasfernog:
LGTM? can't wait to implement this on the wallet lib :D

PR #168, opened 2020-11-19
Enhance the return object of retry()/reattach()/promote() and the description

bingyanglin:
- Now the retry()/reattach()/promote() will return the tuple with type `(MessageId, Message)`.
- Add more description and explanation for the retry() API.

### DIFF
--- a/specs/iota-rs-ENGINEERING-SPEC-0000.md
+++ b/specs/iota-rs-ENGINEERING-SPEC-0000.md
@@ -255,8 +255,7 @@ Following are the steps for implementing this method:
 
 ## `retry()`
 
-Retries (promotes or reattaches) a message for provided message id. Message should only be retried only if they are valid and haven't been
-confirmed for a while. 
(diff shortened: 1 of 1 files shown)


specs/iota-rs-ENGINEERING-SPEC-0000.md
```diff
@@ -255,8 +255,7 @@ Following are the steps for implementing this method:
 
 ## `retry()`
 
-Retries (promotes or reattaches) a message for provided message id. Message should only be retried only if they are valid and haven't been
-confirmed for a while. 
+Retries (promotes or reattaches) a message for provided message id. Message should only be retried only if they are valid and haven't been confirmed for a while. Note that a user needs to maintain a monitoring list of the newly reattached `MessageId`, e.g., if message M1 is reattached by using a new message M2, then the user needs to monitor M1 and M2 before using the retry() again. If either M1 or M2 is already confirmed by the node, then the message should not be retried again. The possibility of using this function is quite low, because the confirmation throughput of the node is expected to be quite high.
```
laumair: Very minor but it should be:
 
Message**s** should only be retried only if they are valid and haven't been confirmed for a while.

laumair (approved):

specs/iota-rs-ENGINEERING-SPEC-0000.md
bingyanglin: Good point. Let me modify.

PR #169, opened 2020-11-19
Update iota-rs-ENGINEERING-SPEC-0000.md

fijter:
Updated the spec with mqtt event wrapper, additional timeout features, default values in arguments, removed the quorum functionality for now, and other small improvements.

### DIFF
--- a/specs/iota-rs-ENGINEERING-SPEC-0000.md
+++ b/specs/iota-rs-ENGINEERING-SPEC-0000.md
@@ -19,6 +19,8 @@ Specification of High Level Abstraction API
   * [`get_balance`](#get_balance)
   * [`get_address_balances`](#get_address_balances)
   * [`retry`](#retry)
+  * [`subscribe`](#subscribe)
+  * [`unsubscribe`](#unsubscribe)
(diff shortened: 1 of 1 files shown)


specs/iota-rs-ENGINEERING-SPEC-0000.md
```diff
@@ -85,18 +101,11 @@ for node in node_pool_urls{
 }
 ```
 
-When a return value from an API (e.g., the balance in an address) needs to be determined by quorum (i.e., not by a single healthy node only), then all the nodes in the `synced_node_list` will be queried to determine the final value.
-
 * Unsolved Questions
```
bingyanglin: I think we can remove the `Unsolved Questions` here, because we already decided to use a background worker to periodically update the `synced_node_list`.

bingyanglin (changes requested):

specs/iota-rs-ENGINEERING-SPEC-0000.md
fijter: Yes, absolutely right :)

bingyanglin (approved):
LGTM

PR #172, opened 2020-11-20
bee-p to master

kwek20:
Whoeps i forgot the client change, only did core in last pr

### DIFF
--- a/iota-client/Cargo.toml
+++ b/iota-client/Cargo.toml
@@ -13,7 +13,7 @@ license = "Apache-2.0"
 name = "iota_client"
 
 [dependencies]
-bee-transaction = { git = "https://github.com/Alex6323/bee-p.git" }
+bee-transaction = { git = "https://github.com/Alex6323/bee-p.git", branch = "master" }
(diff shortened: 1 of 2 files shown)


nothingismagick:
Didn't @thibault-martinez say something about NOT using anything from the bee prototype?

kwek20:
Probably fine for the legacy branch

thibault-martinez:
Why are we spending time on `iota-1.0` though ?
`bee-p/master` hasn't received any update in a long time and won't receive anymore at all.

kwek20:
streams still uses iota-1.0, as i assume all other projects for pre-chrysalis2.0 :=) @thibault-martinez 
No time was needed to be spend tho, its jsut since the default branch changed, we need to specify.

Ill update it to the exact commit instead, so its always good to go

thibault-martinez:
Yeah sure, I am just saying that `bee-p/master` never reached and will never reach maturity. I am not exactly sure what you expect from it, I just want to warn that we won't provide support or invest time in it anymore :)

lucasfernog:
We also won't invest time on iota-1.0 anymore but if this is just to fix some issue on streams until they migrate to chrysalis, that's fine. I'll merge this.

kwek20:
Yea was jsut a warning saying 2 imports using different branches, all good! Thanks @lucasfernog

PR #173, opened 2020-11-23
feat (bindings) Introduce Python binding

bingyanglin:
# IOTA Client Python Library

## Requirements
- Rust 1.45.0+
- Python 3.6+

## Try Run w/ Local Hornet

1. Run your local Hornet
- `$ git clone git@github.com:gohornet/hornet.git`
- checkout `chrysalis-pt2` branch
- Modify your `create_snapshot_alphanet.sh`, modify Line 14 to `go run ../main.go tool snapgen alphanet1 96f9de0989e77d0e150e850a5a600e83045fa57419eaf3b20225b763d4e23813 snapshots/alphanet1/full_export.bin`
- `$ ./run_coo_bootstrap.sh `

2. Build the iota-client-python library by yourself (Optional)
- **IMPORTANT**: Enable the `bindings/python/native` in your `Cargo.toml`
```toml
[workspace]
members = [
    "iota-core",
    "bindings/python/native",
    "iota-client",
    "examples",
]
``` 
- `$ cargo build --release`
- The built library is located in `target/release/`
- On MacOS, rename `libiota_client.dylib` to `iota_client.so`, on Windows `libiota_client.dll` to `iota_client.pyd`, and on Linux `libiota_client.so` to `iota_client.so`.
- Copy your renamed library to `bindings/python/examples/`
- Go to `binding/python/`
- `$ python example.py`

## Python Example
```python
import iota_client
import os
LOCAL_NODE_URL = "http://0.0.0.0:14265"

# NOTE! Load the seed from your env path instead
# NEVER assign the seed directly in your codes!
# DO NOT USE THIS!!:
# SEED = "256a818b2aac458941f7274985a410e57fb750f3a3a67969ece5bd9ae7eef5b2"

# USE THIS INSTEAD
SEED = os.getenv('MY_IOTA_SEED')

EMPTY_ADDRESS = "iot1qxgamuxntdxq06q4zpmvmdnrerj2f94058ge3flfyx567unw25amvr978uw"
client = iota_client.Client(
    node=LOCAL_NODE_URL, node_sync_disabled=True)


def main():
    print('get_health()')
    print(f'health: client.get_health()')

    print('get_info()')
    print(f'node_info: client.get_info()')

    print('get_tips()')
    print(f'tips: client.get_tips()')

    print('find_addresses')
    address_changed_list = client.find_addresses(
        seed=SEED, account_index=0, begin=0, end=10, get_all=True)
    print(f'address_changed list: {address_changed_list}')

    # Get the (address, changed ) for the first found address
    address, changed = address_changed_list[0]
    print(f'get_address_balance() for address {address}')
    print(f'balance: {client.get_address_balance(address)}')

    print(f'get_address_balance() for address {EMPTY_ADDRESS}')
    print(f'balance: {client.get_address_balance(EMPTY_ADDRESS)}')

    print(f'get_address_outputs() for address {EMPTY_ADDRESS}')
    print(f'outputs(): {client.get_address_outputs(EMPTY_ADDRESS)}')

    print(f'send() 100 tokens to address {EMPTY_ADDRESS}')
    message_id = client.send(
        seed=SEED, outputs=[{'address': EMPTY_ADDRESS, 'amount': 100}])
    print(f'Token sent with message_id: {message_id}')
    print(f'Please check http://127.0.0.1:14265/api/v1/messages/{message_id}')

    print(f'get_message_metadata() for message_id {message_id}')
    message_metadata = client.get_message_metadata(message_id)
    print(f'message_metadata: {message_metadata}')

    print(f'get_message_data() for message_id {message_id}')
    message_data = client.get_message_data(message_id)
    print(f'message_data: {message_data}')

    print(f'get_message_raw() for message_id {message_id}')
    message_raw = client.get_message_raw(message_id)
    print(f"message_raw: {bytearray(message_raw, 'utf-8')}")

    print(f'get_message_children() for message_id {message_id}')
    children = client.get_message_children(message_id)
    print(f"children: {children}")

    print(f'send() Indexation')
    message_id_indexation = client.send(
        index="Hello", data=bytes("Tangle", 'utf-8'))
    print(f'Indexation sent with message_id: {message_id_indexation}')
    print(
        f'Please check http://127.0.0.1:14265/api/v1/messages/{message_id_indexation}')

    print(f"get_message_index() for index 'Hello'")
    message_id_indexation_queried = client.get_message_index("Hello")
    print(f'Indexation: {message_id_indexation_queried}')

    print(f"find_messages() for indexation_keys = ['Hello']")
    messages = client.find_messages(indexation_keys=["Hello"])
    print(f'Messages: {messages}')

    print(f"get_unspent_address()")
    unspent_addresses = client.get_unspent_address(seed=SEED)
    print(f'(unspent_address, index): {unspent_addresses}')

    print(f"get_balance()")
    balance = client.get_balance(seed=SEED)
    print(f'balance: {balance}')

    addresses = []
    for address, _changed in address_changed_list:
        addresses.append(address)
    print(f"get_address_balances() for {addresses}")
    balances = client.get_address_balances(addresses)
    print(f'balances: {balance}')


if __name__ == "__main__":
    main()
```


lucasfernog:
`iota-core-python` should probably be moved to `bindings/python`.

lucasfernog:
also I wonder if we should write a C binding and use that to create the Python binding instead, since we'll also need the C binding for Java/Go if we want those bindings.

bingyanglin:
In the python binding we need to add the implementation of every types those expose to the users, like `Address` and others. So we will need to modify all the used crates (including bee crates) to expose them in Python. Can we avoid this if we write C binding and use that for Python binding. Or we might need to implement and maintain the binding whenever the iota.rs is updated.

lucasfernog:
modifying stuff outside iota.rs is overkill and probably wouldn't even be accepted in Bee. We'd need to write a wrapper Address type on the Python binding crate and use that instead. (this also applies to a C binding).

lucasfernog:
We had this initial work for the C binding: https://github.com/iotaledger/iota.rs/tree/iota-1.0/bindings/c

binding/python/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Python Example
+```python
+import iota_client as iota
+
+# Create your client instance
+client = iota.Client("http://0.0.0.0:14265")
+
+# Send your token
+message_id = client.send(seed="256a818b2aac458941f7274985a410e57fb750f3a3a67969ece5bd9ae7eef5b2",
```
nothingismagick: I am opposed to using this type of example where the seed is passed in as a string / shown in cleartext and think we can use this as an opportunity to remind people about the importance of protecting the seed...

bingyanglin: Agree. Would you please provide a example of avoiding this kind of thing. Thank you.

lucasfernog:
For instance, for the Address type on wallet.rs we represent it as a bech32 string to passs around between the binding and the Rust code.

binding/python/README.md
nothingismagick: As discussed in our call yesterday, will discuss this with the stronghold team and make a .env example as well as at least one version of using a CLI stronghold invocation to get the seed.

bingyanglin: Amazing! Thanks a lot.

nothingismagick:
I sort of think that JUST making a C binding is better, because we get Python and Go pretty much for free.

bingyanglin:
@lucasfernog Got it. Thank you for pointing this out.

bingyanglin:
@nothingismagick I think this is a good way to maintain different bindings.

bingyanglin:
> I sort of think that JUST making a C binding is better, because we get Python and Go pretty much for free.

After digging into the python binding and c binding. I think for python part it will be easier/faster/more efficient to just bind the rust from python. An example is that python lacks of enum, which needs some tricky design to bind this. To avoid too many type casting, I binded the rust to python directly, instead of rust -> c -> python. Ofc we can process on the C binding in the future.

bingyanglin:
> also I wonder if we should write a C binding and use that to create the Python binding instead, since we'll also need the C binding for Java/Go if we want those bindings.

After digging into the python binding and c binding. I think for python part it will be easier/faster/more efficient to just bind the rust from python. An example is that python lacks of enum, which needs some tricky design to bind this. To avoid too many type casting, I binded the rust to python directly, instead of rust -> c -> python. Ofc we can process on the C binding in the future.

Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [workspace]
 members = [
     "iota-core",
+    # "bindings/python/native",
```
Thoralf-M: Maybe we should create an extra workspace in the cargo.toml in the python folder like we have for the node binding now?

Thoralf-M:
There are two `iota_client.so` files, one should be enough?

bindings/python/native/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
```
Thoralf-M, line 3: I would add [workspace] above here so we don't need it in the root cargo.toml of the repo

bindings/python/native/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+[dependencies]
+tokio = { version = "1.0", features = ["macros"] }
+hex = "0.4"
+iota-core = { path = "../../../iota-core" }
+syn = "=1.0.57"
+dict_derive = "0.3.0"
+serde_json = "1.0"
+
+[dependencies.pyo3]
+version = "0.12.3"
```
Thoralf-M: For windows we need
version = "0.13"
features = ["extension-module", "abi3"]

bindings/python/native/src/client/high_level_api.rs
```diff
@@ -0,0 +1,285 @@
...
+        network_id: Option<u64>,
+    ) -> Result<Message> {
+        let mut send_builder = self.client.send();
+        if let Some(account_index) = account_index {
+            send_builder = send_builder.with_account_index(account_index);
+        }
+        if let Some(initial_address_index) = initial_address_index {
+            send_builder = send_builder.with_initial_address_index(initial_address_index);
+        }
+        if let Some(inputs) = inputs {
```
Thoralf-M, line 50: Custom range for the inputs is missing

bindings/python/README.md
```diff
@@ -0,0 +1,133 @@
...
+members = [
+    "iota-core",
+    "bindings/python/native",
+    "iota-client",
+    "examples",
+]
+``` 
+- `$ cargo build --release`
+- The built library is located in `target/release/`
+- On MacOS, rename `libiota_client.dylib` to `iota_client.so`, on Windows `libiota_client.dll` to `iota_client.pyd`, and on Linux `libiota_client.so` to `iota_client.so`.
```
Thoralf-M: On Windows the file is only iota_client.dll called, without `lib` before

bindings/python/native/Cargo.toml
bingyanglin: Thank you for pointing this out! Modified.

bindings/python/native/Cargo.toml
bingyanglin: Agree. Modified.

bindings/python/README.md
bingyanglin: Thank you for pointing this out! Modified.

bindings/python/native/src/client/high_level_api.rs
bingyanglin: Thank you for identifying this. Added error handling [here](https://github.com/iotaledger/iota.rs/blob/c1aeb5324724b9ce83ef4f6072c7403d9803a597/bindings/python/native/src/client/high_level_api.rs#L36-L42) and supported user inputs [here](https://github.com/iotaledger/iota.rs/blob/c1aeb5324724b9ce83ef4f6072c7403d9803a597/bindings/python/native/src/client/high_level_api.rs#L58-L60)

Thoralf-M (approved):

PR #189, opened 2020-12-04
feat(client): add node syncing process

lucasfernog:
Introduces a node syncing thread that checks the status of the nodes and adds only healthy nodes to the pool list.


iota-client/src/error.rs
```diff
@@ -49,7 +49,7 @@ impl fmt::Display for Error {
             Error::MissingParameter(s) => write!(f, "Must provide required parameter:{}", s),
             Error::InvalidParameter(s) => write!(f, "Parameter is invalid:{}", s),
             Error::SpentAddress => "Found Spent Address that still has balance.".fmt(f),
-            Error::NodePoolEmpty => "No node available".fmt(f),
+            Error::SyncedNodePoolEmpty => "No node available".fmt(f),
```
thibault-martinez, line 55: https://github.com/dtolnay/thiserror is nice to implement Display on errors :)

iota-client/src/client.rs
```diff
@@ -81,24 +87,26 @@ impl BrokerOptions {
 
 /// An instance of the client using IRI URI
 pub struct Client {
-    /// Node pool of IOTA nodes
-    pub(crate) pool: Arc<RwLock<HashSet<Url>>>,
-    pub(crate) sync: Arc<RwLock<Vec<Url>>>,
+    /// Node pool of synced IOTA nodes
+    pub(crate) sync: Arc<RwLock<HashSet<Url>>>,
+    /// Flag to stop the node syncing
+    pub(crate) stop_sync: Arc<AtomicBool>,
```
thibault-martinez: Maybe prefer a oneshot channel if you want instantaneous stop.

iota-client/src/client.rs
```diff
@@ -108,37 +116,72 @@ impl std::fmt::Debug for Client {
...
+    pub(crate) fn start_sync_process(
+        mut runtime: Runtime,
+        sync: Arc<RwLock<HashSet<Url>>>,
+        nodes: Vec<Url>,
+        node_sync_interval: NonZeroU64,
+        stop_sync: Arc<AtomicBool>,
+    ) -> thread::JoinHandle<()> {
+        let node_sync_interval = Duration::from_millis(node_sync_interval.into());
+
+        thread::spawn(move || {
```
thibault-martinez: I see you have a runtime, any reason not to spawn a task here ?

iota-client/src/client.rs
```diff
@@ -108,37 +116,72 @@ impl std::fmt::Debug for Client {
...
-    //         };
-    //     }
+    pub(crate) fn sync_nodes(
+        runtime: &mut Runtime,
+        sync: &Arc<RwLock<HashSet<Url>>>,
+        nodes: &[Url],
+    ) {
+        let mut synced_nodes = HashSet::new();
+
+        runtime.block_on(async {
```
thibault-martinez: We're sure we want to block on this, right ?

iota-client/src/client.rs
```diff
@@ -108,37 +116,72 @@ impl std::fmt::Debug for Client {
...
 
-    /// Get a node candidate from the node pool.
+    /// Get a node candidate from the synced node pool.
     pub(crate) fn get_node(&self) -> Result<Url> {
-        Ok(self
-            .pool
-            .read()
-            .unwrap()
+        let pool = self.sync.read().unwrap();
+        Ok(pool
```
thibault-martinez: This always return the same peer ? Is this wanted ?

iota-client/src/error.rs
lucasfernog: yeah I'm planning on making this change soon, just need to find some time to improve the overall quality of this crate.

iota-client/src/client.rs
lucasfernog: yeah this will change with the quorum implementation, for now I guess we'll stick with the first node.

iota-client/src/client.rs
lucasfernog: how would that make it instantaneous? do you have an example?

iota-client/src/client.rs
lucasfernog: i'm not sure how do I spawn a periodic task without blocking the current thread

iota-client/src/client.rs
lucasfernog: yeah it's important specially on the first run

iota-client/src/client.rs
thibault-martinez: At the moment, in the worst case scenario, you have to wait a full sleep and then check if the flag has been changed. You can `select` on both a shutdown channel and a ticker and act accordingly. An example here https://github.com/iotaledger/bee/blob/chrysalis-pt-2/bee-protocol/src/worker/status.rs#L37 We use what we call a ShutdownStream to combine a oneshot and a stream but you don't necessarily have to use it, a select would be sufficient.

thibault-martinez: Of course, if potentially waiting a full sleep is not an issue for the client, you can ignore me.

iota-client/src/client.rs
thibault-martinez: I see what you mean. I would have to dive into the whole code base, blocking the main thread with a runtime is usually not an issue if you spawn all you tasks before.

iota-client/src/client.rs
lucasfernog: please see [`3d6501b` (#189)](https://github.com/iotaledger/iota.rs/pull/189/commits/3d6501b82490db40f145b4068b738586c2ef3ca2)

iota-client/src/client.rs
lucasfernog: please see [`3d6501b` (#189)](https://github.com/iotaledger/iota.rs/pull/189/commits/3d6501b82490db40f145b4068b738586c2ef3ca2)

iota-client/src/builder.rs
```diff
@@ -128,21 +128,27 @@ impl ClientBuilder {
         let sync = Arc::new(RwLock::new(HashSet::new()));
         let sync_ = sync.clone();
 
-        let stop_sync = Arc::new(AtomicBool::new(false));
-        let stop_sync_ = stop_sync.clone();
+        let (sync_kill_sender, sync_kill_receiver) = channel(1);
```
thibault-martinez, line 138: I am not sure it really needs to be a `broadcast`, `oneshot` if made for this purpose and probably sufficient ?

iota-client/src/client.rs
```diff
@@ -134,42 +137,45 @@ impl Client {
...
+                tokio::select! {
+                    _ = async {
+                        loop {
+                            // delay first since the first `sync_nodes` call is made by the builder
+                            // to ensure the node list is filled before the client is used
+                            delay_for(node_sync_interval).await;
+                            Client::sync_nodes(&sync, &nodes).await;
+                        }
+                    } => {}
+                    _ = kill.recv() => {}
```
thibault-martinez: Have you had the chance to try this ? I am having a hard time figuring out if this is correct or not

thibault-martinez: I would have looped over a select but this may be right as well, a bit unusual though

lucasfernog: yeah that was my first try but it wasn't compiling :P but now I got it to work

iota-client/src/builder.rs
lucasfernog: I tried with an oneshot channel but I think I didn't understand its usage with tokio::select. First of all the try_recv method isn't async so the behavior is different. Then it also panics saying that I read it after it was complete.

PR #190, opened 2020-12-04
feat(client): implement local PoW, closes #91

### LINKED ISSUE #91: Local POW is missing
It seems like we lost the local POW functionality that was present in the old `iota-lib-rs`. Will it be reimplemented?

lucasfernog:


thibault-martinez (approved):
LGTM

PR #196, opened 2020-12-10
Allow sending to own address

Thoralf-M:


Thoralf-M:
I had one output with 2779530283277761 i on an address and wanted to send 1000 to the same address again, that didn't work, because the library added the address also as a second output, which isn't allowed `Error: IotaClientError(MessageError(DuplicateError))` 
https://github.com/iotaledger/bee/blob/e0bc13b9d1de595e89a3a1e67c60ffe4e26c52b0/bee-message/src/payload/transaction/essence.rs#L194

The new `new_address_list` vector is `output_addresses` with the addresses from the outputs removed, so that there can't be 2 outputs with the same address anymore, if there are funds left and a new output is created

PR #202, opened 2020-12-14
convenient indexation message builder, removal of quorum code, making examples/tests compile since of changes in Bee

samuel-rufi:
# Description of change

This PR includes an convenient builder for Indexation messages. This builder can be used in a similar fashion as the already existing transaction builder. An example of how to use that builder is provided.
Furthermore, this PR removes quorum code (including removal of quorum code in the bindings) since not required by the spec.
Tried to fix broken examples and tests (were broken since of recent changes in Bee).

## Type of change

- Bug fix

The changes were tested against the alphanet (Hornet)
Furthermore I did self-review my own code.

## Change checklist

- [X] I have followed the contribution guidelines for this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have checked that new and existing unit tests pass locally with my changes
- [X] I have updated the CHANGELOG.md, if my changes are significant enough

### DIFF
--- a/iota-client/src/client.rs
+++ b/iota-client/src/client.rs
@@ -479,7 +479,7 @@ impl Client {
     }
 
     /// A generic send function for easily sending value transaction messages.
-    pub fn send_indexation_message<'a>(&'a self) -> SendIndexationBuilder<'a> {
+    pub fn send_indexation_message(&self) -> SendIndexationBuilder<'_> {
         SendIndexationBuilder::new(self)
     }
 
--- a/iota-client/src/client.rs
+++ b/iota-client/src/client.rs
@@ -478,8 +478,13 @@ impl Client {
     //////////////////////////////////////////////////////////////////////
 
     /// A generic send function for easily sending value transaction messages.
-    pub fn send<'a>(&'a self, seed: &'a Seed) -> SendBuilder<'a> {
-        SendBuilder::new(self, seed)
+    pub fn send_transaction_message<'a>(&'a self, seed: &'a Seed) -> SendTransactionBuilder<'a> {
+        SendTransactionBuilder::new(self, seed)
+    }
+
+    /// A generic send function for easily sending value transaction messages.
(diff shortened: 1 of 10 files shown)


lucasfernog:
Overall this is a very good PR, nice work!

The only thing that I'd like to check is the `send` entry point on client.rs. I think we should have only one `send` function exposed, and keep the `SendBuilder` struct. That struct would then have a `transaction` and `indexation` methods, which would redirect to the actual `SendTransactionBuilder` and `SendIndexationBuilder`.

cc @fijter

fijter:
Thanks Samuel! I fully agree with @lucasfernog to have 1 send method and use the builder. It should be possible to send transactions with value and without, where a indexationPayload should be either part of the message (data only) or as part of the payload of the signedTransaction (value tx).

iota-client/src/client.rs
```diff
@@ -478,8 +474,13 @@ impl Client {
     //////////////////////////////////////////////////////////////////////
 
     /// A generic send function for easily sending value transaction messages.
-    pub fn send<'a>(&'a self, seed: &'a Seed) -> SendBuilder<'a> {
-        SendBuilder::new(self, seed)
+    pub fn send_transaction_message<'a>(&'a self, seed: &'a Seed) -> SendTransactionBuilder<'a> {
+        SendTransactionBuilder::new(self, seed)
+    }
+
+    /// A generic send function for easily sending value transaction messages.
```
Alex6323: fix documentation (c&p)

iota-client/src/client.rs
```diff
@@ -478,8 +474,13 @@ impl Client {
...
+    pub fn send_indexation_message(&self) -> SendIndexationBuilder<'_> {
+        SendIndexationBuilder::new(self)
     }
 
     /// Return a valid unuspent address.
```
Alex6323:
```suggestion
    /// Return a valid unspent address.
```

PR #209, opened 2020-12-15
Allow messages without a payload

Thoralf-M:


lucasfernog:
I saw this issue and wanted to fix it, thanks for doing it first <3

PR #231, opened 2021-01-05
Update spec

Thoralf-M:
[Rendered](https://github.com/iotaledger/iota.rs/blob/update-spec/specs/iota-rs-ENGINEERING-SPEC-0000.md)
Fix #242

PR #240, opened 2021-01-10
refactor sendBuilder, add custom parent

Thoralf-M:


iota-client/tests/node_api.rs
```diff
@@ -73,7 +73,7 @@ async fn test_post_message_with_indexation() {
 
     let r = client
         .send()
-        .indexation("Hello".to_string())
+        .with_index("Hello")
```
bingyanglin, line 76: How do you think we change all the name `index` to be `indexation`, so it is aligned with the full name described in the spec? So here we will have `with_indexation()`

bindings/node/tests/client.js
```diff
@@ -42,7 +42,7 @@ describe('Client', () => {
     const depositAddress = 'iot1q9jyad2efwyq7ldg9u6eqg5krxdqawgcdxvhjlmxrveylrt4fgaqj30s9qj'
     const messageId = await client
       .send()
-      .transaction(seed)
+      .with_seed(seed)
```
bingyanglin, line 45: I like this, much clearer than the previous method naming.

iota-client/tests/node_api.rs
bingyanglin: After thinking twice with_index is better, because indexation has `index` and `data`, which are defined in the bee crates

bingyanglin (commented):
Looks really nice.

iota-client/src/api/send.rs
```diff
@@ -330,69 +358,47 @@ impl<'a> SendTransactionBuilder<'a> {
...
+        self.finish_message(Some(payload)).await
     }
 
     /// Consume the builder and get the API result
-    pub async fn finish(self) -> Result<MessageId> {
-        let index = self.index;
-        let data = self.data.unwrap_or_default();
+    pub async fn finish_indexation(self) -> Result<MessageId> {
+        let index = self.index.clone();
+        let data = self.data.clone().unwrap_or_default();
```
bingyanglin: Why we don't consume them but need to clone them?

Thoralf-M: Because otherwise we can't use self.finish_message() at the end

 ```iota-client\src\api\send.rs:386:9
     |
378  |         let data = self.data.unwrap_or_default();
     |                              ------------------- `self.data` partially moved due to this method call
...
386  |         self.finish_message(Some(payload)).await
     |         ^^^^ value used here after partial move
     |
note: this function consumes the receiver `self` by taking ownership of it, which moves `self.data`
     |
1177 |     pub fn unwrap_or_default(self) -> T {
     |                              ^^^^
     = note: partial move occurs because `self.data` has type `std::option::Option<std::vec::Vec<u8>>`, which does not implement the `Copy` trait```

iota-client/tests/node_api.rs
Thoralf-M: But good point about the spec, I already updated the name in https://github.com/iotaledger/iota.rs/pull/231, but the naming is probably not consistent everywhere yet

iota-client/src/api/send.rs
bingyanglin: But it seems that we don't need the index and data after `finish_indexation()`. In the `finish_message()`, we don't need the `index` and `data` anymore. Am I right? I mean in rust we might be able to design the codes to consume the `index` and `data` in `finish_indexation()`, or we cannot?

bingyanglin: Like the following codes, which can build pass. In this way we can avoid the uncessary clone() of the fields.
```rust
pub async fn finish_indexation(self) -> Result<MessageId> {
    let payload: Payload;
    {
        let index = &self.index.as_ref();
        let empty_slice = &vec![];
        let data = &self.data.as_ref().unwrap_or(empty_slice);

        // build indexation
        let index = Indexation::new(index.expect("No indexation tag").to_string(), data)
            .map_err(|e| Error::IndexationError(e.to_string()))?;
        payload = Payload::Indexation(Box::new(index));
    }

    // building message
    self.finish_message(Some(payload)).await
}
```

Thoralf-M: Is this `let empty_slice = &vec![];` required for this to work or works `unwrap_or_default` too?

bingyanglin: I wanted to use `unwrap_or_default` before. Unfortunately it seems we will need clone() inevitably if we use `unwrap_or_default`

bingyanglin: The error follows
```
error[E0599]: no method named `unwrap_or_default` found for enum `std::option::Option<&std::vec::Vec<u8>>` in the current scope
   --> iota-client/src/api/send.rs:381:44
    |
381 |             let data = &self.data.as_ref().unwrap_or_default();
    |                                            ^^^^^^^^^^^^^^^^^ method not found in `std::option::Option<&std::vec::Vec<u8>>`
    |
    = note: the method `unwrap_or_default` exists but the following trait bounds were not satisfied:
            `&std::vec::Vec<u8>: std::default::Default`
 ```

Thoralf-M: Ok, fine anyway, will you push it or should I do it?

bingyanglin: Pushed.

bingyanglin (approved):

PR #241, opened 2021-01-11
use u64 instead of NonZeroU64

Thoralf-M:


nothingismagick:
If this is going to be exposed anywhere, U64 types are problematic in a number of Apple contexts. Please have @lucasfernog review this before merging.

lucasfernog:
I think we should just go with u64 for now since we're not supporting WASM yet (and we need this change everywhere).

lucasfernog (approved):

PR #247, opened 2021-01-13
use seed from .env for examples

### LINKED ISSUE #232: Use .env for seeds in examples

Thoralf-M:


.env
```diff
@@ -0,0 +1,2 @@
+seed=256a818b2aac458941f7274985a410e57fb750f3a3a67969ece5bd9ae7eef5b2
```
rajivshah3: To remove this from Git, run `git rm -r --cached .`, and then `git add .` and `git commit`

.github/workflows/build_and_test.yml
```diff
@@ -75,10 +75,12 @@ jobs:
     # paho.mqtt requires openssl and OPENSSL_DIR on Windows
     - name: Install OpenSSL (Windows)
       if: matrix.os == 'windows-latest'
-      run: |
```
rajivshah3: The `|` pipe is necessary because there are multiple commands running in this step

Thoralf-M: Just noticed that, thanks

bingyanglin (approved):

PR #250, opened 2021-01-13
use f64 instead of usize and store network related info locally

Thoralf-M:


iota-client/src/client.rs
```diff
@@ -232,7 +232,7 @@ impl std::fmt::Debug for Client {
         d.field("sync", &self.sync).field("client", &self.client);
         #[cfg(feature = "mqtt")]
         d.field("broker_options", &self.broker_options);
-        d.field("local_pow", &self.local_pow).finish()
+        d.field("local_pow", &self.network_info.local_pow).finish()
```
lucasfernog: should we dump the whole network_info object here instead of just the local_pow field?

iota-client/src/client.rs
```diff
@@ -292,21 +291,22 @@ impl Client {
...
-                    if network == Network::Testnet && info.network_id == "mainnet"
-                        || network == Network::Mainnet && info.network_id != "mainnet"
+                    if network_info.read().unwrap().network == Network::Testnet && info.network_id == "mainnet"
+                        || network_info.read().unwrap().network == Network::Mainnet && info.network_id != "mainnet"
                     {
                         continue;
                     }
-                    if !local_pow {
+                    let mut client_network_info = network_info.write().unwrap();
+                    client_network_info.min_pow_score = info.min_pow_score;
```
lucasfernog, line 308: should we also set `network_info.local_pow` to true if the node doesn't have the PoW feature?

Thoralf-M: No, then the node should not be selected

iota-client/src/builder.rs
```diff
@@ -169,7 +184,7 @@ impl ClientBuilder {
             mqtt_topic_handlers: Default::default(),
             #[cfg(feature = "mqtt")]
             broker_options: self.broker_options,
-            local_pow: self.local_pow,
+            network_info: network_info.read().unwrap().clone(),
```
lucasfernog: you're updating the `network_info` on the Client::sync_nodes function but it won't reflect the changes on the field here; you should store the Arc<RwLock<NetworkInfo>> instead

iota-client/src/client.rs
Thoralf-M: Makes sense

iota-client/src/builder.rs
Thoralf-M: Thanks, updated

iota-client/src/client.rs
Thoralf-M: At least that's what I think, because local_pow is true as default and if you disable it you should have a reason for it, maybe running on a device with a slow cpu, automatically enabling it could cause problems or make the message useless, if the PoW takes too long

lucasfernog (approved):

PR #262, opened 2021-01-17
Implement dust protection

Thoralf-M:
Check if a transaction would result with dust on an address, if that's the case, return an error


Thoralf-M:
Still need to check if we use dust allowance outputs as inputs, that no dust will be left which can't be there
Bindings need to be updated

iota-client/src/api/send.rs
```diff
@@ -101,6 +106,19 @@ impl<'a> SendBuilder<'a> {
...
 
+    /// Set a dust allowance transfer to the builder, address needs to be Bech32 encoded
+    pub fn with_dust_allowance_output(mut self, address: &Bech32Address, amount: u64) -> Result<Self> {
+        if amount < 1_000_000 {
+            return Err(Error::DustError(
+                "Amount for SignatureLockedDustAllowanceOutput needs to be >= 1_000_000".into(),
+            ));
+        }
+        let address = Address::try_from_bech32(&address.to_string())?;
+        let output = SignatureLockedDustAllowanceOutput::new(address, amount).unwrap().into();
```
bingyanglin: Am thinking that it might be better to transmit the error out by `?` instead of unwarp() which will panic when there is an error.

iota-client/src/api/send.rs
```diff
@@ -533,25 +527,25 @@ impl<'a> SendBuilder<'a> {
...
+    let mut dust_outputs_amount: i64 = 0;
 
     // add outputs from this transaction
     for output in outputs {
         match output.2 {
             // add newly created outputs
             true => {
                 if output.0 >= 1_000_000 {
-                    dust_allowance_balance += output.0;
+                    dust_allowance_balance += output.0 as i64;
```
bingyanglin, line 543: Can I know that we need u64 or i64 here?

bingyanglin (approved):

PR #270, opened 2021-01-21
feat(client): add `get_peers` API

lucasfernog:


Thoralf-M:
We should use the types from here https://github.com/iotaledger/bee/blob/161f7498e7d52fd6c205b4cbd511aa8754eadad8/bee-api/bee-rest-api/src/types.rs#L12

bernardoaraujor:
we need `get_peers()` for  the Rosetta implementation too.

Thoralf-M (approved):

PR #272, opened 2021-01-24
Fix input selection and replace dotenv crate

Thoralf-M:


Thoralf-M:
Fix is only `if total_already_spent >= total_to_spend {`, the rest is to replace the dotenv crate

iota-client/src/api/send.rs
```diff
@@ -297,7 +301,7 @@ impl<'a> SendBuilder<'a> {
                                 }
                             }
                         }
-                        if total_already_spent > total_to_spend {
+                        if total_already_spent >= total_to_spend {
```
bingyanglin, line 304: Note that if `total_already_spent == total_to_spend`, then we don't have to put `0` tokens back to the original address, i.e., we should use `>` in this condition checking.

bingyanglin (changes requested):

iota-client/src/api/send.rs
Thoralf-M: But we have this in line 290?

bingyanglin: Both should be `>` I think. We don't have to put signature when there is no tokens being transferred.

Thoralf-M: What do you mean with signature?
Currently with 
```Rust
if total_already_spent > total_to_spend {
                            break 'input_selection;
                        }
```
it will continue to search for inputs even if the amount == the amount we want to spend, so if we want to spend 10 Mi and have one output with 10 Mi it will still continue to search for other inputs
If we change it to `total_already_spent >= total_to_spend` it will break from the loop with the single 10 Mi output which is enough, no 0 output will be added (not even possible) and there will also not be another signature (which would also make it invalid)

bingyanglin: Got it. Good point. Then we use this to speed it up. And remain Lines 290/304 to be `>`, which is already there. Thanks!

bingyanglin (approved):

PR #279, opened 2021-01-26
add storage adapter

### LINKED ISSUE #275: Implement state adapter

Thoralf-M:
Fixes https://github.com/iotaledger/iota.rs/issues/275


iota-client/src/storage/account.rs
```diff
@@ -0,0 +1,188 @@
+// Copyright 2020 IOTA Stiftung
```
nothingismagick: 2021

nothingismagick:
Kind of wondering if it might make sense to also offer Stronghold as a storage adapter here...

Thoralf-M:
@nothingismagick what benefit do you expect from this? Because we only store the messages we sent to the Tangle and are public anyways (at least I made it this way now, don't know if it would be useful to store more)

bingyanglin (approved):

PR #283, opened 2021-02-01
fix(client): post message timeout

lucasfernog:


iota-client/src/api/send.rs
```diff
@@ -520,7 +520,8 @@ impl<'a> SendBuilder<'a> {
             .finish()
             .map_err(Error::MessageError)?;
 
-        self.client.post_message(&final_message).await?;
+        let id = self.client.post_message(&final_message).await?;
+        println!("{:?} {:?}", id, final_message.id().0);
         Ok(final_message)
```
Thoralf-M: Remove this println

iota-client/src/client.rs
```diff
@@ -311,6 +311,10 @@ impl Client {
                     }
                     let mut client_network_info = network_info.write().unwrap();
                     client_network_info.network_id = hash_network(&info.network_id);
+                    println!(
```
Thoralf-M: And here

Thoralf-M (approved):

PR #289, opened 2021-02-02
Fix address usage, export message id in MessageWrapper and update tests

### LINKED ISSUE #286: fix nodejs tests

Thoralf-M:
milestone metadata still needs to be fixed and we need an option to get the message id from a message (wrapper around the message object, function to calculate it or both)


Thoralf-M:
Fixes #286

bindings/nodejs/tests/client.js
```diff
@@ -2,12 +2,11 @@ const { ClientBuilder } = require('../lib')
 const { assertAddress, assertMessageId, assertMessage } = require('./assertions')
 const assert = require('assert')
 
-const seed = 'b3a9bf35521157aa9c4508ab3a9266e210ae297ff5a4584234c4d9e7d01712e3'
```
lucasfernog, line 5: we should probably use a env here in the future

lucasfernog (approved):

PR #298, opened 2021-02-05
add peers example

bernardoaraujor:


bernardoaraujor:
@rufsam I think we need to fix something on `bee`.

from @Thoralf-M 's words:

> we are using the types from bee https://github.com/iotaledger/bee/blob/337d74652cf25468cc02ce9c0eedc69d19709678/bee-api/bee-rest-api/src/types.rs#L607
> And the gossip field is also here https://editor.swagger.io/?url=https://raw.githubusercontent.com/rufsam/protocol-rfcs/master/text/0026-rest-api/rest-api.yaml
> So looks like it just wasn't added it hornet yet

examples/peers.rs
```diff
@@ -0,0 +1,20 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! cargo run --example nodeinfo --release
+use iota::Client;
+
+/// In this example we get the nodeinfo
+#[tokio::main]
+async fn main() {
+    let iota = Client::builder() // Crate a client instance builder
+        .with_network("alphanet1")
```
Thoralf-M: Please remove with_network, it's not needed and could cause problems

bernardoaraujor: you mean only line 11, correct?

Thoralf-M: Yes

Thoralf-M (approved):

examples/peers.rs
```diff
@@ -0,0 +1,19 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! cargo run --example nodeinfo --release
```
Thoralf-M: Oh and update this to `//! cargo run --example peers --release`

PR #312, opened 2021-02-10
Reworked get_network_info api

### LINKED ISSUE #310: The `get_network_info` API should fail if no synced node are available
On the wallet.rs i'm using the `get_network_info` API to get the bech32 HRP for my addresses. When the node isn't synced, it returns the default NetworkInfo object. IMO that's wrong, because it leads me to save invalid addresses for the network.

Maybe the default object should instead be `None`, filled by the `sync` process and by the `get_network_info` API itself if it finds that the info is empty.

melatron:
* This is related to issue #310 
* Added get_synced_network_info function which makes sure that the
  network_info object is synced.
* Removed unnecessary use of get_network_info where clone wasn't needed,
  now uses direct getters to the specific fields.
* Changed get_network_id to use  get_synced_network_info

### DIFF
--- a/iota-client/src/client.rs
+++ b/iota-client/src/client.rs
@@ -393,7 +393,7 @@ impl Client {
     /// Gets the miner to use based on the PoW setting
     pub fn get_pow_provider(&self) -> ClientMiner {
         ClientMinerBuilder::new()
-            .with_local_pow(self.network_info.read().unwrap().local_pow)
+            .with_local_pow(self.get_local_pow())
             .finish()
     }
 
@@ -564,7 +564,7 @@ impl Client {
         let mut url = self.get_node()?;
         url.set_path("api/v1/messages");
 
-        let timeout = if self.network_info.read().unwrap().local_pow {
+        let timeout = if self.get_local_pow() {
             self.get_timeout(Api::PostMessage)
         } else {
             self.get_timeout(Api::PostMessageWithRemotePow)
--- a/iota-client/src/client.rs
+++ b/iota-client/src/client.rs
@@ -380,14 +380,8 @@ impl Client {
 
     /// Gets the network id of the node we're connecting to.
     pub async fn get_network_id(&self) -> Result<u64> {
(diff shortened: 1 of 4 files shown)


iota-client/src/client.rs
```diff
@@ -396,10 +393,39 @@ impl Client {
...
+        self.network_info.read().unwrap().tips_interval
+    }
+
+    /// returns the local pow
+       pub fn get_local_pow(&self) -> bool {
+        self.network_info.read().unwrap().local_pow
+    }
+
+    /// returns the min pow score
+    pub fn get_min_pow_score(&self) -> f64 {
```
Thoralf-M: Didn't we want to use it also only with the synced network info?

iota-client/src/client.rs
```diff
@@ -396,10 +393,39 @@ impl Client {
...
+        self.network_info.read().unwrap().min_pow_score
+    }
+
     /// Gets the network related information such as network_id and min_pow_score
     pub fn get_network_info(&self) -> NetworkInfo {
```
Thoralf-M: Maybe we should remove get_network_info? Because for the values that aren't dependent on the node you added new values and here someone should rather use the synced network_info

Thoralf-M:
In the nodejs binding we call `let info = client.get_network_info();`, https://github.com/iotaledger/iota.rs/blob/9859c3348573a89464382676c5459f116f386541/bindings/nodejs/native/src/classes/client/mod.rs#L209
that should also be updated to use the synced version

iota-client/src/client.rs
```diff
@@ -383,21 +383,50 @@ impl Client {
         let network_id = match self.get_network_info().network_id {
```
Thoralf-M: Speaks anything against just using
```Rust
let network_info = self.get_synced_network_info().await?;
Ok(network_info.network_id.unwrap())
```
?

Thoralf-M (approved):

PR #315, opened 2021-02-11
only sign the hashed_essence

### LINKED ISSUE #314: Sign essence hash instead of whole essence

Thoralf-M:


Thoralf-M:
We should wait before we merge it, so transactions on the current testnet will still work

bingyanglin (approved):

PR #324, opened 2021-02-15
Update balance example

Thoralf-M:
Use index 0 as default, update example and fix deadlock

### DIFF
--- a/examples/balance.rs
+++ b/examples/balance.rs
@@ -2,7 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //! cargo run --example balance --release
-use iota::Client;
+use iota::{Client, Seed};
+extern crate dotenv;
+use dotenv::dotenv;
+use std::env;
 
 /// In this example we will get the balance of a known address
 #[tokio::main]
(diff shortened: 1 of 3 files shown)


examples/balance.rs
```diff
@@ -25,4 +28,11 @@ async fn main() {
 
     let output = iota.get_output(&outputs[0]).await.unwrap();
     println!("Output {:?}", output);
+
+    println!("This example uses dotenv, which is not safe for use in production.");
+    dotenv().ok();
+    let seed =
+        Seed::from_bytes(&hex::decode(env::var("NONSECURE_USE_OF_DEVELOPMENT_SEED_1").unwrap()).unwrap()).unwrap();
+    let seed_balance = iota.get_balance(&seed).finish().await.unwrap();
+    println!("Account balance: {:?}i", seed_balance);
```
melatron, line 37: Is the "i" needed inside the println?

Thoralf-M: It's the unit, that's why I added it

melatron (approved):

PR #330, opened 2021-02-16
use specific commits for dependencies and use hash method from bee

### LINKED ISSUE #326: Use specific commit for dependencies
We should always use specific commits so updated in bee doesn't break iota.rs and everything that is dependent on it

Thoralf-M:


melatron (approved):
What is the reason to use specific commits instead of branch?

nothingismagick:
This is a temporary fix because bee isn't ready for the signing revision.

Issue #327, opened 2021-02-15
Have optional logs for API requests and responses


Thoralf-M:

melatron:
Fixed with #346

PR #346, opened 2021-02-22
Added simple request logging macro

melatron:
Added a log_request macro which logs the Request method and URL and the Response status code.


melatron:
Related to #327

Thoralf-M (approved):

PR #359, opened 2021-02-24
Add python binding wheel build flow

### LINKED ISSUE #362: Provide cross-platform prebuilt python wheels
Add workflow to enable users download the pre-built python wheels directly

bingyanglin:
Close #362


.github/workflows/python_binding_publish.yml
```diff
@@ -0,0 +1,94 @@
+name: Build python binding wheels
+
+on:
+  push:
+    branches:
+      - dev
+      - master
+  pull_request:
```
Thoralf-M: pull_request was only added for testing and should be removed now?

bingyanglin: Yes agree. We can remain this only for `push` to speed up the workflow.

Thoralf-M (approved):

PR #373, opened 2021-03-02
Replace reqwest with ureq

### LINKED ISSUE #370: Replace reqwest with ureq
We should add support for basic auth at the same time

Thoralf-M:


Thoralf-M:
Post message doesn't work properly yet `Response[status: 308, status_text: Permanent Redirect]`

Thoralf-M:
> Post message doesn't work properly yet `Response[status: 308, status_text: Permanent Redirect]`

Happens only when using "http" instead of "https" for the testnet nodes

iota-client/src/node/message.rs
```diff
@@ -59,18 +58,16 @@ impl<'a> GetMessageBuilder<'a> {
...
-            let meta = resp.json::<MessagesWrapper>().await?;
-            Ok(
-                Message::try_from(&meta.data).expect("Can't convert MessageDto to Message"))
-        })
+        let resp: ResponseWrapper = get_ureq_agent(self.client.get_timeout(Api::GetMessage))
+            .get(&url.to_string())
+            .call()?
+            .into_json()?;
+
+        Ok(Message::try_from(&resp.data).expect("Can't convert MessageDto to Message"))
```
lucasfernog: we shouldn't panic here

lucasfernog (approved):

PR #406, opened 2021-03-09
refactory (Python binding): Shared runtime and acquire GIL after other computation

### LINKED ISSUE #410: (Python binding) Consider obtaining GIL after serialising event to minimise critical section
GIL should be acquired after computation to reduce the blocking time.

bingyanglin:
Fix #410 #411


bingyanglin:
@zesterer would you please help review this PR?

PR #407, opened 2021-03-09
bug fix: Fix crackability bug and set the init crackability to be f64::MAX

bingyanglin:


Thoralf-M:
I will resolve the conflicts

PR #416, opened 2021-03-10
Sort parents, inputs and outputs

### LINKED ISSUE #397: Sort all parents, inputs and outputs
We need the `lexicographical order of their serialized form`
https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md

Thoralf-M:


Thoralf-M:
@lucasfernog before wallet.rs updates to this or a later commit we also need to have the sorting there, otherwise it will probably break since the sorting in bee is also removed with this commit

lucasfernog (approved):

PR #422, opened 2021-03-15
validate that urls use http or https

### LINKED ISSUE #404: Verify URLs
Something like `Url::parse("data:text/plain,Hello?World#")?;` should return an error, but it doesn't at the moment

Thoralf-M:
Closes #404


melatron (approved):

nothingismagick (approved):
Thanks for sorting this concern out. One less boundary to worry about.

PR #454, opened 2021-03-25
Don't create multiple remainder outputs

### LINKED ISSUE #451: Multiple outputs for multiple custom inputs shouldn't create multiple outputs for the same address
MessageError(DuplicateError) happens when we have more inputs than we need from the same address and therefore create multiple remainder outputs

Thoralf-M:
Fixes #451

### DIFF
--- a/bindings/nodejs/lib/index.d.ts
+++ b/bindings/nodejs/lib/index.d.ts
@@ -99,7 +99,7 @@ export declare class Client {
   getAddressBalance(address: string): Promise<AddressBalance>
   isAddressValid(address: string): boolean
   getMilestone(index: number): Promise<MilestoneMetadata>
-  GetMilestoneUtxoChanges(index: number): Promise<MilestoneUTXOChanges>
+  getMilestoneUtxoChanges(index: number): Promise<MilestoneUTXOChanges>
   getReceipts(): Promise<Receipts[]>
   getReceiptsMigratedAt(index: number): Promise<Receipts[]>
   getTreasury(): Promise<Treasury>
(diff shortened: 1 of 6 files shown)


bindings/nodejs/lib/index.d.ts
```diff
@@ -99,7 +99,7 @@ export declare class Client {
   getAddressBalance(address: string): Promise<AddressBalance>
   isAddressValid(address: string): boolean
   getMilestone(index: number): Promise<MilestoneMetadata>
-  getMilestoneUTXOChanges(index: number): Promise<MilestoneUTXOChanges>
+  GetMilestoneUtxoChanges(index: number): Promise<MilestoneUTXOChanges>
```
bingyanglin: A popular convention in Javascript is to only capitalize constructors, for methods the first world should be lowercase, i.e., `getMilestoneUtxoChanges`

bindings/nodejs/native/src/classes/client/mod.rs
```diff
@@ -470,7 +470,7 @@ declare_types! {
             Ok(cx.undefined().upcast())
         }
 
-        method getMilestoneUTXOChanges(mut cx) {
+        method GetMilestoneUtxoChanges(mut cx) {
```
bingyanglin: Same as the above.

bindings/nodejs/lib/index.d.ts
Thoralf-M: Yeah, but Clippy complained about it, should I add some ignore rule?

bingyanglin: The Clippy complained about `UTXO`, which should be `Utxo`. the `get` prefix should be OK.

Thoralf-M: Oh right, sry read it wrong

bingyanglin: No no don't be. This PR is quite good.

bingyanglin (approved):

PR #457, opened 2021-03-28
Node manager

### LINKED ISSUE #445: Add node url to the get_info() endpoint
## Description

Add `node` property to the response object of `get_info()`. 

## Motivation

Since iota.rs supports multiple nodes, the node it actually queries under the hood is unknown to the caller. It would be helpful if along with the info `node` url is also returned as part of the response object. That way, it will be helpful for the caller to know which node was actually queried for info. 

## Requirements

Add `node` property to the response object with the URL of the node that was queried. 

## Open questions (optional)

N/A

No.

### LINKED ISSUE #446: Quorum
## Description

Quorum for some API endpoints (balance, outputs), do we also need it for other endpoints?

## Motivation

With that we can be more sure that the result is correct

Thoralf-M:
Add a node manager to handle the requests with optional primary node, primary pow node and quorum
Also updates get_info(), so the url of the node is returned together with the nodeinfo


melatron (approved):

bindings/python/examples/example.py
```diff
@@ -15,7 +15,7 @@
 
 EMPTY_ADDRESS = "atoi1qzt0nhsf38nh6rs4p6zs5knqp6psgha9wsv74uajqgjmwc75ugupx3y7x0r"
 client = iota_client.Client(
-    node=LOCAL_NODE_URL, node_sync_disabled=True)
+    nodes_name_password=[[LOCAL_NODE_URL]], node_sync_disabled=True)
```
lucasfernog, line 18: is this change right?

Thoralf-M: For Python @bingyanglin decided with me that we only use a single field there, with which you can do the same stuff, so we don't need multiple fields

iota-client/Cargo.toml
```diff
@@ -19,28 +19,34 @@ bee-pow = { git = "https://github.com/iotaledger/bee.git", rev = "e8cafb6591ec0c
...
 once_cell = { version = "1.5", optional = true }
-futures = { version = "0.3", optional = true }
+
+#storage
+async-trait = "0.1"
+rusqlite = { version = "0.24", features = ["bundled"], optional = true }
 
 [features]
-default = ["async"]
+default = ["async", "futures"]
```
lucasfernog, line 48: is `futures` dependency here still used? couldn't find it

Thoralf-M: It's used here https://github.com/iotaledger/iota.rs/blob/9d44e7ccc0b9dc8f8392e9d717f3af4a8d893846/iota-client/src/node_manager.rs#L133

lucasfernog: `futures` seems like a required dependency, not an optional one

lucasfernog (approved):

PR #464, opened 2021-03-31
improve is_dust_allowed

### LINKED ISSUE #461: Improve dust check
## Description

The API has a `dustAllowed` field https://api.hornet-0.testnet.chrysalis2.com/api/v1/addresses/atoi1qzt0nhsf38nh6rs4p6zs5knqp6psgha9wsv74uajqgjmwc75ugupx3y7x0r
With that we don't need to fetch all outputs from the addresses here https://github.com/iotaledger/iota.rs/blob/59da05593e4a16a4d11b219782828ed37c68c38c/iota-client/src/api/message_builder.rs#L576, but can just use this field to determine if we can send dust to this address

## Motivation

Better performance when dust or dust allowance outputs are involved

If nobody else creates one before.

Thoralf-M:


Thoralf-M:
Dependent on https://github.com/iotaledger/bee/issues/496

bingyanglin (approved):

PR #474, opened 2021-04-07
Updated to latest bee revs

melatron:


Thoralf-M:
Updated it in the nodeinfo pr

PR #476, opened 2021-04-07
return address bech32 encoded for balance and outputs

Thoralf-M:
Python binding needs to be updated the same way


bingyanglin:
Docs need to be changed too.

bingyanglin (approved):

PR #477, opened 2021-04-08
(feat) Python binding: upload python wheels to release tags by workflow

bingyanglin:


bingyanglin:
Close this issue because the workflow for uploading cross-platform wheels to pypi is integrated in this [repo](https://github.com/bingyanglin/iota.rs/tree/maturin_publish)

PR #479, opened 2021-04-08
use default as default feature

Thoralf-M:


iota-core/Cargo.toml
```diff
@@ -15,7 +15,7 @@ name = "iota"
 iota-client = { version = "0.5.0-alpha", path = "../iota-client", default-features = false }
```
lucasfernog, line 15: remove =false here

iota-core/Cargo.toml
```diff
@@ -15,7 +15,7 @@ name = "iota"
...
 
 [features]
-default = ["iota-client/async"]
+default = ["iota-client/default"]
```
lucasfernog, line 18: and then you don't need this line

lucasfernog (changes requested):

lucasfernog (approved):

PR #494, opened 2021-04-12
Remove IOTA 1.0 notes from Docs

Hribek25:
- IOTA 1.0 comparisons were removed and replaced with an overal chapter at Chrysalis Docs

### DIFF
--- a/docs/libraries/nodejs/README.md
+++ b/docs/libraries/nodejs/README.md
@@ -11,7 +11,7 @@ This library is in active development. The library targets the Chrysalis testnet
 If you want to get involved in discussions about this library, or you're looking for support, go to the #clients-discussion channel on [Discord](https://discord.iota.org).
 
 ## Changes from IOTA 1.0 to 1.5
-For an overview of all significant changes in the protocol, please see [Chrysalis documentation](https://chrysalis.docs.iota.org/guides/index.html#overall-changes-from-iota-10-to-15-chrysalis-in-a-nutshell).
+For an overview of all significant changes in the protocol, please see [Chrysalis documentation](https://chrysalis.docs.iota.org/guides/index.html).
 
 ## Overview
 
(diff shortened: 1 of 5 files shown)


docs/libraries/nodejs/README.md
```diff
@@ -10,6 +10,9 @@ This library is in active development. The library targets the Chrysalis testnet
 
 If you want to get involved in discussions about this library, or you're looking for support, go to the #clients-discussion channel on [Discord](https://discord.iota.org).
 
+## Changes from IOTA 1.0 to 1.5
+For an overview of all significant changes in the protocol, please see [Chrysalis documentation](https://chrysalis.docs.iota.org/guides/index.html#overall-changes-from-iota-10-to-15-chrysalis-in-a-nutshell).
```
Thoralf-M: Can this be `from-iota-1-to-1(5)` or similar? 10 to 15 could be confusing

Hribek25: Unfo it is not possible but I solved it somehow. I realized that we can refer to top level page instead of the specific chapter. It will be more stable moving forward...

docs/welcome.md
```diff
@@ -7,59 +7,14 @@ Example of tasks that `iota.rs` is able to help with:
...
-            <td colspan=4>iota = mainnet; atoi = testnet</td>
-        </tr>
-    </tbody>
-</table>
-
-More details: [RFC: Bech32 Address Format](https://github.com/iotaledger/protocol-rfcs/pull/20)
+> Please note: there is also available `wallet.rs` library that contains all the logic to safely build wallets or integrations that require value-based IOTA transfers. We strongly recommend to leverage [wallet.rs library](https://wallet-lib.docs.iota.org/) in case one is more focused on IOTA value-based transfers since it integrates the best security practices including our [stronghold enclave](https://blog.iota.org/iota-stronghold-6ce55d311d7c/).
 
+## Changes from IOTA 1.0 to 1.5
+For an overview of all significant changes in the protocol, please see [Chrysalis documentation](https://chrysalis.docs.iota.org/guides/index.html#overall-changes-from-iota-10-to-15-chrysalis-in-a-nutshell).
```
Thoralf-M: Maybe the same here so it's consistent?

Hribek25: Good catch. Thank you!

Thoralf-M (approved):

PR #495, opened 2021-04-12
[Docs] Change API reference link to link to iota-lib-rs docs

monospacedmagic:
This is purely a documentation fix, or rather a suggestion thereof.

I noticed that the API reference link in the project's readme file links to the [iota-core docs](https://docs.rs/iota-core/), however, from what I can tell iota-core is just a subcrate of this project.

That's why I suggested to change the API reference URL to the [iota-lib-rs docs](https://docs.rs/iota-lib-rs/), as they seem to document this project as a whole.

Please let me know if this change is correct. Thank you for reading!

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ When using the "MQTT" feature, connecting to a MQTT broker using raw ip doesn't 
 
 ## API reference
 
-You can read the [API reference](https://docs.rs/iota-core) here, or generate it yourself.
+You can read the [API reference](https://docs.rs/iota-lib-rs/) here, or generate it yourself.
 
 If you'd like to explore the implementation in more depth, the following command generates docs for the whole crate, including private modules:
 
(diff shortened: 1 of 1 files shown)


Thoralf-M:
Thanks for the PR, unfortunately the docs there are still for the iota 1.0 library because the new one didn't get published to crates.io yet
But we want to update the link to docs not on crates.io in the next days

README.md
```diff
@@ -51,7 +51,7 @@ When using the "MQTT" feature, connecting to a MQTT broker using raw ip doesn't
 
 ## API reference
 
-You can read the [API reference](https://docs.rs/iota-core) here, or generate it yourself.
+You can read the [API reference](https://docs.rs/iota-lib-rs/) here, or generate it yourself.
```
huhn511: You can use "https://client-lib.docs.iota.org/docs/doc/iota/index.html" and fix #497 with that :)

Thoralf-M (approved):

PR #500, opened 2021-04-14
add mqtt example

Thoralf-M:


bindings/nodejs/examples/10_mqtt.js
```diff
@@ -3,8 +3,10 @@ async function run() {
         ClientBuilder
     } = require('@iota/client');
 
-    // client will connect to testnet by default
-    const client = new ClientBuilder().build();
+    // client connects to a node that has MQTT enabled
+    const client = new ClientBuilder()
+        .node('https://api.hornet-0.testnet.chrysalis2.com')
```
Hribek25, line 8: I think we need to use a specific node that is MQTT-enabled and not a load balancer, right?

Based on https://chrysalis.docs.iota.org/testnet.html

Hribek25 (approved):
- Added MQTT chapter to Docs
- slightly updated example to use a specific node that is MQTT enabled, right?

bindings/nodejs/examples/10_mqtt.js
Thoralf-M: I tried it and it worked since we don't only add the load balancer but all testnet nodes, but yeah specifying the node should be better for this example

PR #533, opened 2021-05-03
nodejs/fix hexToBech32 and update versions

### LINKED ISSUE #530: Nodejs binding - fix timeout function name and add supported nodejs version
There is `#### defaultTimeout(timeoutMs): ClientBuilder` which we don't have, but we have `requestTimeout()` and we should add the supported nodejs versions

### LINKED ISSUE #532: NodeJS: Client::hexToBech32 fails if HRP is not provided
## Bug description

An exception is thrown if `Client::hexToBech32` is called without a second argument.

## Version

- @iota/client": "^0.3.1" (from npmjs.org)

## Steps To reproduce the bug

1.  Compile and run the following valid TS script:
```
import { ClientBuilder } from '@iota/client'

async function run() {
    try {
        const client = new ClientBuilder().build()
        const address = "iota1qqadr8ee669rx0475c0pe7upwsvyh5p7x5tq3kt8ucrv55h8ythqx5r750c"
        let hexAddress = client.bech32ToHex(address)
        let bech32Address = await client.hexToBech32(hexAddress, "iota")
        console.log(bech32Address, hexAddress)
    } catch(e) {
        console.log(e)
    }
}
run()
```

2.  Remove the "iota" second argument to `hexToBech32`
3.  Script will throw an exception:
```
TypeError: failed to downcast any to string
    at /node_modules/@iota/client/lib/index.js:17:48
    at new Promise (<anonymous>)
    at ClientWrapper.hexToBech32 (/node_modules/@iota/client/lib/index.js:17:12)
    at run (/dist/test.js:9:42)
    at Object.<anonymous> (/dist/test.js:16:1)
    at Module._compile (internal/modules/cjs/loader.js:1137:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1157:10)
    at Module.load (internal/modules/cjs/loader.js:985:32)
    at Function.Module._load (internal/modules/cjs/loader.js:878:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
```

## Expected behaviour

You can:
- change the signature of the function and make the second argument non-optional or
- have the function works even if the second argument is absent

## Actual behaviour

Exception thrown if the second argument is not here as the signature suggests.

Thoralf-M:


bingyanglin (approved):

coinwalletdev:
Thanks!

PR #552, opened 2021-05-08
Fix address type

gson88:
The type for address was `'string'` (with quotation-marks), instead of just `string`, which could result in unnecessary TS errors.


Thoralf-M (approved):
Thanks for the PR!

Issue #560, opened 2021-05-14
Automatically encode non hex inputs for MQTT indexation to hex


Thoralf-M:
## Description

If the index is not hex encoded we should do it

Comment here should be removed then again https://github.com/iotaledger/iota.rs/commit/54a7380784ee4abaa7bf792c156eabefaca1a630

## Motivation

Easier to use

## Are you planning to do it yourself in a pull request?

Maybe.

Issue #524, opened 2021-04-28
Support MQTT over websockets


Thoralf-M:
## Description

Support MQTT over websockets, it's supported from rumqttc now https://docs.rs/rumqttc/0.5.0/rumqttc/enum.Transport.html

## Motivation

With that no additional TCP port is needed for the nodes

PR #561, opened 2021-05-14
Mqtt websocket first iteration

melatron:


melatron:
Closes #524

Thoralf-M:
Closes https://github.com/iotaledger/iota.rs/issues/560

iota-client/src/node/mqtt.rs
```diff
@@ -83,9 +96,19 @@ async fn get_mqtt_client(client: &mut Client) -> Result<&mut MqttClient> {
...
                 let host = node.host_str().expect("Can't get host from URL");
-                let mut mqttoptions = MqttOptions::new(host, host, 1883);
-                mqttoptions.set_connection_timeout(client.broker_options.timeout.as_secs());
-                let (_, mut connection) = MqttClient::new(mqttoptions.clone(), 10);
+                let id = "iota.rs";
+                let mut port = client.broker_options.port;
+                let mut uri = format!("wss://{}:{}/mqtt", host, node.port_or_known_default().unwrap_or(port));
+                if !client.broker_options.use_ws {
+                    uri = host.to_string();
+                    port = 1883;
```
Thoralf-M: I think `port = 1883;` should be removed now, because it's the default port above and if the user provides another one we should also use the one from the user provided

iota-client/src/client.rs
```diff
@@ -114,6 +114,10 @@ pub struct BrokerOptions {
     pub(crate) automatic_disconnect: bool,
     #[serde(default = "default_broker_timeout")]
     pub(crate) timeout: Duration,
+    #[serde(default = "default_broker_use_ws")]
```
Thoralf-M: We should also use `rename = "defaultBrokerUseWs"` here, because in nodejs it's expected to be camelCase

iota-client/src/client.rs
```diff
@@ -114,6 +114,10 @@ pub struct BrokerOptions {
...
+    pub(crate) use_ws: bool,
+    #[serde(default = "default_broker_port")]
```
Thoralf-M: Same as above

Thoralf-M (approved):

PR #588, opened 2021-06-07
Use ubuntu-18.04 for Pythong workflow

Thoralf-M:
```
Hi. I'm trying to get the python iota_client (0.2.0-a8) running on an AWS (Linux 2) (LightSail) server.
It runs fine on my Mac (Catalina) but I'm getting an ImportError failure on first run on AWS: saying the required glibc version 2.29 is not found (AWS has glibc 2.26).
```
We had the same problem with the nodejs binding and building it with ubuntu-18.04 fixed it

### DIFF
--- a/.github/workflows/python_binding_publish.yml
+++ b/.github/workflows/python_binding_publish.yml
@@ -9,7 +9,7 @@ on:
 jobs:
 
   linux-wheels:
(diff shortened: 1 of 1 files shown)


bingyanglin:
Thanks!

bingyanglin (approved):

PR #589, opened 2021-06-10
Support offline signing

Thoralf-M:


kwek20:
@Thoralf-M LGTM! Only thing is maybe we should add the `PreparedTransactionData` to iota.rs somewhere as well. and then instead of returning the result tuple of essence and AddressIndexRecorder, just return that.
Why did you choose to leave that up to the user?

Thoralf-M:
> @Thoralf-M LGTM! Only thing is maybe we should add the `PreparedTransactionData` to iota.rs somewhere as well. and then instead of returning the result tuple of essence and AddressIndexRecorder, just return that.
> Why did you choose to leave that up to the user?

Didn't thought that much about it and just used the structure that was already there from the offline stuff, but good point, will change it

kwek20 (approved):

melatron (approved):

PR #603, opened 2021-07-05
(feat) Python binding: add prepare/sign_transaction functions

bingyanglin:
- Added `prepare_transaction` Python binding
- Added `sign_transaction` Python binding
- Small refactory of type casting between Python and Rust

### DIFF
--- a/iota-client/Cargo.toml
+++ b/iota-client/Cargo.toml
@@ -18,7 +18,8 @@ bee-rest-api = "0.1"
 bee-message = "0.1"
 bee-pow = "0.1"
 bee-common = "0.4"
-iota-crypto = { version = "0.5", features = ["std", "blake2b", "ed25519", "random", "slip10", "bip39", "bip39-en"]}
+# TODO: Use "0.6" when the iota-crypto is updated.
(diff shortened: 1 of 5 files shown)


iota-client/Cargo.toml
```diff
@@ -18,7 +18,8 @@ bee-rest-api = "0.1"
 bee-message = "0.1"
 bee-pow = "0.1"
 bee-common = "0.4"
-iota-crypto = { version = "0.5", features = ["std", "blake2b", "ed25519", "random", "slip10", "bip39", "bip39-en"]}
+# TODO: Use "0.6" when the iota-crypto is updated.
+iota-crypto = { git = "https://github.com/iotaledger/crypto.rs", features = ["std", "blake2b", "ed25519", "random", "slip10", "bip39", "bip39-en"], rev = "e42be4d"}
```
Thoralf-M, line 22: iota-crypto got updated now

Thoralf-M (approved):

iota-client/Cargo.toml
bingyanglin: Yes but it's only for 0.6.0, we will need the `e42be4d` to be included in their next updated version.

PR #630, opened 2021-07-27
Java/docs: Getting started and api docs

kwek20:
Docs:
- Getting started for Gradle and Maven
- API reference (made by hand :shivers: )
- Typos fixed in generated java_gue file
-
Bindings:
- Fixed that we now return byte[] instead of short[]
- Added More Display/toString methods
- Correct return types and some more getters in types

### DIFF
--- a/documentation/docs/libraries/java/api_reference.md
+++ b/documentation/docs/libraries/java/api_reference.md
@@ -1,0 +1,11 @@
+# API Reference
+
+API reference
+
+You can read the [API reference](https://docs.rs/iota-client) here, or generate it yourself.
+
+If you'd like to explore the implementation in more depth, the following command generates docs for the whole crate, including private modules:
+
+```bash
+cargo doc --document-private-items --no-deps --open
+```
--- a/documentation/docs/libraries/java/api_reference.md
+++ b/documentation/docs/libraries/java/api_reference.md
@@ -1,11 +1,1096 @@
-# API Reference
+# API Reference - IOTA Client Library - Java binding
 
-API reference
+### Api
 
-You can read the [API reference](https://docs.rs/iota-client) here, or generate it yourself.
+The Api enum contains a list of node endpoints we can call.
+It is used in setting a timeout for a specific api call during the buiding of a client using `withApiTimeout(api, timeout)` .  
 
-If you'd like to explore the implementation in more depth, the following command generates docs for the whole crate, including private modules:
+GET_HEALTH
+GET_INFO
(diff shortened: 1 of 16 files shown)


kwek20:
@Thoralf-M Whats up with that docosaurus error? 🤔

documentation/docs/libraries/java/troubleshooting.md
```diff
@@ -0,0 +1,11 @@
+# Troubleshooting
+
+
+## StackExchange
+
+> [https://iota.stackexchange.com](https://iota.stackexchange.com/)
```
Thoralf-M, line 6: @kwek20 maybe this `>` is the problem, but also not sure

Thoralf-M: @lucas-tortora do you maybe know what exactly is the issue, could it be this `>`?

documentation/docs/libraries/java/api_reference.md
```diff
@@ -1,11 +1,1096 @@
...
+
+Returns the sum of all `MigratedFundsEntry` items within a `ReceiptPayload`.
+
+#### transaction(): TreasuryPayload
+
+The `TreasuryTransaction` used to fund the funds of a `ReceiptPayload`.
+
+### Essence
+
+#### getAsRegular(): Optional<RegularEssence>
```
lucas-tortora: @Thoralf-M , @kwek20 The Docusaurus build error comes from these tags.  Docusuaurus interprets them as HTML tags which are not closed.  There's 2 options to work around this:

1. Close the "tag": `<RegularEssence/>`, though I'm not sure if that's correct syntax for JAVA
2. Use &lt; and &gt; to replace < and >: `&lt;RegularEssence&gt;`

documentation/docs/libraries/java/troubleshooting.md
lucas-tortora: @kwek20 , since we're using docusarus now, you can use [admonitions](https://docusaurus.io/docs/next/markdown-features/admonitions) if you want.

documentation/docs/libraries/java/api_reference.md
kwek20: I will do option 2. One just looks really weird :)

kwek20:
@Thoralf-M Ready to go :D Clippy warnings are in iota.rs itself,. not bindings

Thoralf-M (approved):

Issue #47, opened 2020-05-03
get_node_info doesn't work with hornet


Thoralf-M:
```RUST
use iota_client;

fn main() {
  println!("{:#?}", iota_client::get_node_info("https://nodes.comnet.thetangle.org".to_string()).unwrap());
}
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error(Json(Error("missing field `appName`", line: 1, column: 80)))', C:\Users\Boss\.cargo\registry\src\github.com-1ecc6299db9ec823\iota-client-0.1.0\lib.rs:272:44

wusyong:
I tried to test on a hornet node and it worked. Could you test on get-node-info example with your node? (Or DM yours, I can help testing.)

```
GetNodeInfoResponse {
    app_name: "HORNET",
    app_version: "0.4.0-rc8",
    jre_available_processors: None,
    jre_free_memory: None,
    jre_max_memory: None,
    jre_total_memory: None,
    jre_version: None,
    latest_milestone: "HF9ZIOOGOPZLQSEIRUHMTFMBCCQTGOLHFUCSHDEDAOUURFUFSDXKOKPNNEJGSCRAOIUQRQPUYPVGZ9999",
    latest_milestone_index: 1399405,
    latest_solid_subtangle_milestone: "HF9ZIOOGOPZLQSEIRUHMTFMBCCQTGOLHFUCSHDEDAOUURFUFSDXKOKPNNEJGSCRAOIUQRQPUYPVGZ9999",
    latest_solid_subtangle_milestone_index: 1399405,
    milestone_start_index: 1392635,
    neighbors: 7,
    packets_queue_size: None,
    time: 1588517203000,
    tips: 0,
    transactions_to_request: 0,
}
```

wusyong:
Ah just note that new update is not available on crates.io yet. Since bee crates haven't pushed to there. I'm unable to push either. You might want to import current library with git dependency.
```
iota-client = { git = "https://github.com/iotaledger/iota.rs" }
```

Thoralf-M:
https://github.com/wusyong/iota-example/blob/master/src/main.rs works with the new version

PR #631, opened 2021-07-28
Wasm binding

### LINKED ISSUE #555: Browser-support for iota.rs / WASM-bindings
## Description

Currently, iota.rs does not work in the browser, which makes it unusable for (existing legacy-) webapps. As webapps build a common use case for the client-libs, creating wasm-bindings would make a lot of sense. I hesitate to use the iota.js-library, as it most likely will not be maintained to the same degree as iota.rs.

## Motivation

This would be very useful for webapps that require the client lib in any shape or form.

## Open questions

- If wasm-bindings are already planned, it would be great if a timeline would be shared so that we can allign it with our Chrysalis-migration-strategy. 
- Otherwise, a clear plan for the future of iota.js would be great and whether it is recommended to be used in longterm-projects

No.

### LINKED ISSUE #658: Improve node error message
Current error isn't very clear

Thoralf-M:
Closes https://github.com/iotaledger/iota.rs/issues/555
# Wasm Binding Check List

## Builder
- Check list for all parameters
    -  network
    -  primary_node
    -  primary_pow_node
    -  permanode
    -  node_auth
    -  node
    -  nodes
    -  node_sync_interval
    -  node_sync_disabled
    -  quorum
    -  quorum_size
    -  quorum_threshold
    -  node_pool_urls
    -  offline_mode
    - [x] update spec
    -  local_pow
    -  request_time_out
    -  broker_options
    -  api_timeout

## General API
- [x] `message()`
-  seed
-  data
-  account_index 
-  initial_address_index
-  input
-  outputs
-  index
-  parents

- [x] `get_message()`
-  metadata()
-  data()
-  raw()
-  children()
-  index

- [x] `find_messages()`
-  indexation_key
-  message_ids

- [x] `get_unspent_address()`
-  seed
-  account_index 
-  initial_address_index 

- [x] `get_balance()`
-  seed
-  account_index 
-  initial_address_index 

- [x] `get_address_balances()`
-  addresses

- [x] `is_address_valid()`
-  address

- [x] `generate_mnemonic()`

- [x] `mnemonic_to_hex_seed()`
-  mnemonic

- [x] `retry()`
-  message_id

- [x] `retry_until_included()`
-  message_id
-  interval
-  max_attempts

- [x] `reattach()`
-  message_id

- [x] `promote()`
-  message_id

- [x] `bech32_to_hex()`
-  address

- [x] `hex_to_bech32()`
-  address
-  bech32_hrp

- [x] `consolidate_funds()`
-  seed
-  account_index
-  address_range

## Full Node API
- [x] `get_health()`

- [x] `get_info()`

- [x] `get_tips()`

- [x] `get_peers()`

- [x] `post_message()`

- [x] `get_output()`
-  outputId

- [x] `get_address()`
-  address

- [x] `find_outputs()`
-  output_id
-  addresses

- [x] `get_milestone()`
-  index

- [x] `get_milestone_utxo_changes()`
-  milestone_index

- [x] `get_receipts()`

- [x] `get_receipts_migrated_at()`
-  milestone_index

- [x] `get_treasury()`

- [x] `get_included_message()`
-  message_id


kwek20:
Thanks! Works with streams wasm bindings as well again now :=)

Thoralf-M:
Will look if we can reuse https://github.com/iotaledger/identity.rs/pull/344/files

kwek20:
@Thoralf-M Whats left for this to get merged? Its kind of annoying to keep switching between dev and wasm when testing :)

Thoralf-M:
@kwek20 only a review and possible changes because of that😅

bindings/python/native/tests/test_full_node_api.py
```diff
@@ -64,7 +64,9 @@ def test_get_address_outputs():
 
 def test_find_outputs():
     outputs = client.find_outputs(addresses=[tv['ADDRESS'][0]])
-    assert isinstance(outputs, list) and 'message_id' in outputs[0]
+    assert isinstance(outputs, list)
+    if len(outputs) > 0:
+        'message_id' in outputs[0]
```
melatron: Is this doing something? Maybe we need to add an assert there?

bindings/wasm/native/build/clean.js
```diff
@@ -0,0 +1,6 @@
+const fs = require('fs')
+const path = require('path')
+const nodefile = path.join(__dirname, '../node/.gitignore')
+fs.unlinkSync(nodefile);
+const webfile = path.join(__dirname, '../web/.gitignore')
+fs.unlinkSync(webfile);
```
melatron:
```suggestion
fs.unlinkSync(webfile);

```

bindings/wasm/native/static/index.html
```diff
@@ -0,0 +1,13 @@
...
+<head>
+  <meta charset="UTF-8">
+  <title>IOTA client</title>
+</head>
+
+<body>
+  <script src="index.js"></script>
+</body>
+
+</html>
```
melatron:
```suggestion
</html>

```

bindings/wasm/native/tests/assertions.js
```diff
@@ -0,0 +1,25 @@
...
+  assert.strictEqual(typeof address, 'string')
+  assert.strictEqual(address.length, 64)
+  assert.strictEqual(address.startsWith('atoi'), true)
+}
+
+module.exports = {
+  assertMessageWrapper,
+  assertMessageId,
+  assertAddress
+}
```
melatron:
```suggestion
}

```

bindings/wasm/native/tests/client.js
```diff
@@ -0,0 +1,183 @@
...
+        .finishMessage(signed_transaction);
+    })
+
+    it('public key to address', async () => {
+      const address = await client.hexPublicKeyToBech32Address("2baaf3bca8ace9f862e60184bd3e79df25ff230f7eaaa4c7f03daa9833ba854a")
+      assert.strictEqual(address, 'atoi1qzt0nhsf38nh6rs4p6zs5knqp6psgha9wsv74uajqgjmwc75ugupx3y7x0r')
+    })
+  })
+}
+test()
```
melatron:
```suggestion
test()

```

melatron (approved):

bindings/python/native/tests/test_full_node_api.py
Thoralf-M: Good catch, yes it should have an assert there

bindings/wasm/native/build/node.js
```diff
@@ -0,0 +1,20 @@
...
+// Add node fetch stuff (https://github.com/seanmonstar/reqwest/issues/910)
+const entryFilePathNode = path.join(__dirname, '../node/iota_client_wasm.js')
+const entryFileNode = fs.readFileSync(entryFilePathNode).toString()
+let changedFileNode = entryFileNode.replace(
+    "let imports = {};",
+    "const fetch = require(\'node-fetch\')\r\nglobal.Headers = fetch.Headers\r\nglobal.Request = fetch.Request\r\nglobal.Response = fetch.Response\r\nglobal.fetch = fetch\r\n\r\nlet imports = {};"
+)
+    // workaround for node.js, find out if that's required only with nodejs 14 or what's the issue there
+    .replace(
+        "getObject(arg0).now()",
```
kwek20: Ive been having env issues related to this fix:  (not caused by)
`WebAssembly.Instance(): Import #47 module="env" function="now" error: function import requires a callable`

Why did you add this workaround? It doesnt seem to help, in my case atleast :)

A quick fix for me was...
```
#[wasm_bindgen]
pub fn now() {}
```

bindings/wasm/native/src/client_builder.rs
```diff
@@ -0,0 +1,265 @@
...
+  #[wasm_bindgen(js_name = apiTimeout)]
+  pub fn api_timeout(&mut self, api: String, timeout: u32) -> Result<ClientBuilder, JsValue> {
+    let api = Api::from_str(&api)
+      .map_err(|_| iota_client::Error::ApiError)
+      .map_err(wasm_error)?;
+    self.with_mut(|builder| builder.with_api_timeout(api.clone(), to_duration(timeout)))?;
+    Ok(self.clone())
+  }
+
+  pub(crate) fn take_builder(&mut self) -> Result<RustClientBuilder, JsValue> {
```
kwek20: Whats up with all this shenanigans? This works fine :)
https://github.com/kwek20/iota.rs/commit/b5be0590dd11e33d256b8fa9829cbcfc0c623ee1

Or is it for some error reason, you catch and can continue the builder? Im not sure if thats really needed, if your builder fails, your program likely does as well

bindings/wasm/native/tests/client.js
```diff
@@ -0,0 +1,183 @@
+const { ClientBuilder } = require('../node')
+const { assertAddress, assertMessageId, assertMessageWrapper } = require('./assertions')
+const assert = require('assert')
+
+const seed = '256a818b2aac458941f7274985a410e57fb750f3a3a67969ece5bd9ae7eef5b2'
+
+async function test() {
+
+  const client = await new ClientBuilder()
+    .node('https://api.thin-hornet-2.h.chrysalis-devnet.iota.cafe/')
```
kwek20: This one aint working at all :D
`failed, reason: getaddrinfo EAI_AGAIN`

bindings/wasm/native/Cargo.toml
```diff
@@ -0,0 +1,38 @@
+[workspace]
+
+[package]
+name = "iota-client-wasm"
```
kwek20: For streams and iota.rs/nodejs we got rid of the iota prefix, as publishing them would already be @iota/client-wasm.

Its at least consistent then. Would require changes to the imports on all examples if you chose to

bindings/wasm/native/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "test-iota-client-wasm",
```
kwek20: And this would be called @iota/cllient-wasm then :D (see my comment on "native/Cargo.toml"

kwek20 (commented):
The URL you use in the examples doesn't work for me. 
`https://api.lb-0.h.chrysalis-devnet.iota.cafe`. But i ran them with mainnet and it works.
Personally i find the error i got back a bit strange! `Error [iota_client::Error]: Response error with status code 404: 404 page not found`

IF the url is wrong, we should replace that across the board, theres like a billion spots :D

bindings/wasm/native/package.json
Thoralf-M: Yeah `@iota/client-wasm` sounds good I think for the release, but need to check with the others

bindings/wasm/native/Cargo.toml
Thoralf-M: Why would it be `@iota/client-wasm` when publishing it? The npm package maybe, but not the Rust crate?

bindings/wasm/native/tests/client.js
Thoralf-M: The 2 in the url was wrong, thanks for catching that

bindings/wasm/native/src/client_builder.rs
Thoralf-M: Thanks for pointing this out, yes I think your approach is better, will merge it

bindings/wasm/native/build/node.js
Thoralf-M: In Nodejs I got `TypeError: Cannot read property 'now' of undefined` when calling a function where a timeout was used `retryUntilIncluded` or `consolidateFunds` (the timeout isn't reached when a message is already confirmed or no funds need to be consolidated)
The wasm-timer crate calls JS for the timeouts
```
#[wasm_bindgen]
pub fn now() {}
``` 
might compile, but will probably fail then

What did you do to get `WebAssembly.Instance(): Import #47 module="env" function="now" error: function import requires a callable`?

Thoralf-M: You can try if it works with 
```JS
    let message = await client.message().submit()
    await client.retryUntilIncluded(message.messageId, BigInt(2), BigInt(5));
```

bindings/python/native/tests/fixtures/test_vectors.json
```diff
@@ -31,7 +31,7 @@
         ]
     },
     "UTXOINPUT": [
-        "7702ea0f2cd6af3206b894c3f2fe4362b23f0f4828857d31e733103b09db25840000"
+        "00000000000000000000000000000000000000000000000000000000000000000000"
```
laumair: What's the purpose for this change here? @Thoralf-M

Thoralf-M: The other output exists only in the old testnet and the nodes of the new devnet don't know about it, but about the genesis output

Thoralf-M: Ok apparently not anymore, but I need funds first so I can create a new output that will not be moved

laumair (approved):
Looks good to me! Great work @Thoralf-M 🎉 

I would suggest to do before release:

- Update version and add alpha tag wherever required;
- In `package.json` in directories, only examples is listed. Perhaps, we should update it to reflect all directories;

Issue #661, opened 2021-08-30
[Task]: Create a workflow for pypi uploading


laumair:
## Description

Create a workflow for pypi uploading.

/cc @bingyanglin

thibault-martinez:
Package is created but not uploaded.

PR #680, opened 2021-09-14
(feat) Python binding: Fix python upload-artifact flow and add pypi upload flow

bingyanglin:
Solved #661 

TODO: Integrate the workflow with the `.change` covector flow.
Notes:
- Now the python wheels will be uploaed to pypi automatically when new codes are pushed to `pip_workflow` branch.
- In pypi we cannot upload the same version twice.
- The wheel version is defined in `bindings/python/native/Cargo.toml`


jyhi:
I'll try adopting this change as a part of the workflow refurbishment I'm working on, as the work can definitely be done in a better way (e.g. using `matrix` instead of copy-pasting Python versions) so this PR might not get merged...

bingyanglin:
Previously we used `matrix` for different os and python versions in the workflow, but there are some slight differences in the commands for them, so in this draft I temperately separate them. Would be quite nice if you can refactory them.

PR #702, opened 2021-09-28
Update Covector workflow for Wasm

Thoralf-M:


Thoralf-M:
wasm-binding-prebuild: can probably removed again, but it shouldn't run the nodejs-binding-prebuild if there is a change in the wasm binding

Thoralf-M:
nodejs-binding-prebuild: and  electron-binding-prebuild is executed also for the wasm-binding, with this additional check I hope that it gets skipped
```
  packagesPublished:
    description: Comma separated list of all of the packages that published. Most useful for chaining runs together to act on the filtered list.
```
https://github.com/jbolda/covector/blob/028ae8967a82c2dd8f8898f682599f5dcadb049d/packages/action/action.yml

laumair (approved):

PR #706, opened 2021-09-30
feat/wasm: Wasm export

kwek20:
We would like to expose the wasm bindings so that we can reference them in streams and not duplicate the builder but just import :=)

Question: Do we expose all modules?


Thoralf-M:
Why do we need to change the workspace for this?

kwek20:
@Thoralf-M Hmm i was under the impression it was needed to import. I just tested it and its not 🙃 
Im baffled as to how rust knows where `client-wasm` is; magic? 

Do you know why it jsut works like this?
`client-wasm = { git = "https://github.com/kwek20/iota.rs", branch  = "wasm-export", default-features = false }`

Also; How do you feel about renaming the package.json from `@iota/client-wasm` to `@iota/client`. Or would that conflict with the JS bindings? We want to rename in streams but also keep consistency across IF :)

Thoralf-M:
@kwek20 
```
Cargo will fetch the git repository at this location then look for a Cargo.toml for the requested crate anywhere inside the git repository (not necessarily at the root - for example, specifying a member crate name of a workspace and setting git to the repository containing the workspace).
```
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

> Or would that conflict with the JS bindings?

Yes, `@iota/client` is used for the nodejs binding already, so we can't use it

kwek20:
Such magic is rust! Thanks for explaining @Thoralf-M 
I fixed it all, and this commit history is now a mess :D

What about my question? Do we expose all modules?

Thoralf-M:
@kwek20 if you think that it could be useful then go for it

Thoralf-M (approved):

kwek20:
Ok did the useful ones! Lets merge!

Thoralf-M:
Thanks!

PR #724, opened 2021-10-27
Add Cargo.lock files

Thoralf-M:
We want to have the lock files so we can be sure that we always use the same dependencies


bingyanglin (approved):
LGTM!

PR #734, opened 2021-11-03
Correct python wheels upload flow

bingyanglin:
- Correct the python wheels upload flow, currently only python3.6 wheels are uploaded in the artifact for different python versions


Thoralf-M (approved):

Thoralf-M:
Sry for breaking all the workflows, should be fixed in https://github.com/iotaledger/iota.rs/pull/733

PR #735, opened 2021-11-03
Update RelationDto

bingyanglin:
Update the RelationDto type in python binding due to the updated bee-rest-api version


bingyanglin:
Duplicated.

PR #741, opened 2021-11-18
Restructure Repository Layout

jyhi:
This PR:

- Pulls `iota-client` out from `iota-client/`, making it a top-level crate
- Tweaks `Cargo.toml` to accommodate the change
- Tweaks `README.md`

... in the hope of making the package look more intuitive.

Bindings continue stay at `bindings/` as separate crates (are they?).


jyhi:
There are failing CIs... too much damage

jyhi:
It should be good now; the failing CIs need another PR to fix.

Thoralf-M:
With this change `cargo run --example 10_mqtt --release` doesn't work anymore, because `mqtt` is not a default feature, could you change the comment in the 10_mqtt.rs example to `cargo run --example 10_mqtt --release --features="mqtt"`?

Cargo.toml
```diff
@@ -1,5 +1,84 @@
...
+zeroize = { version = "1.4.1", default-features = false }
+
+# sync
+ureq = { version = "2.2.0", default-features = false, features = ["tls", "json"], optional = true }
+# async
+reqwest = { version = "0.11.6", default-features = false, features = ["json", "rustls-tls", "blocking"], optional = true }
+futures = { version = "0.3.17", default-features = false, optional = true }
+
+# MQTT
+rumqttc = { version = "0.10", default-features = false, features = ["websocket"], optional = true }
```
Thoralf-M:
```suggestion
rumqttc = { version = "0.10.0", default-features = false, features = ["websocket"], optional = true }
```

Cargo.toml
```diff
@@ -1,5 +1,84 @@
...
+version = "1.1.0"
+authors = [
+  "Thoralf Müller <thoralf.mueller@iota.org",
+  "Antony Dikov <antoni.dikov@iota.org>",
+  "Bing-Yang Lin <bingyang.lin@iota.org>",
+  "Lucas Nogueira <lucas.nogueira@iota.org>",
+  "Nathan Jaremko <nathan@jaremko.ca>",
+  "Yu-Wei Wu <wusyong9104@gmail.com>"
 ]
+edition = "2018"
```
Thoralf-M, line 12: Maybe we should also update to 2021, wdyt?

jyhi: _Do it! Just do it!_ (but in another PR of course)

Thoralf-M (approved):

PR #743, opened 2021-11-20
Fix method name in Java example

safehammad:
Method incorrectly named leading to compilation error when running Java example.

**To reproduce:**

Build and run Java example as described in `iota.rs/bindings/java/README.md`:
```
$ cd iota.rs/bindings/java
$ ./gradlew examples:basic-app:test --info
```

**Resulting error:**

```
Compiling with JDK Java compiler API.
./iota.rs/bindings/java/examples/basic-app/src/main/java/org/example/ExampleApp.java:58: error: cannot find symbol
            System.out.println(RustHex.encode(secret_key.to_bytes()));
                                                        ^
  symbol:   method to_bytes()
  location: variable secret_key of type SecretKey
1 error
```


jyhi (approved):
LGTM.

PR #747, opened 2021-11-23
bindings/java: fix useless use of `format!`

jyhi:
A Clippy run across the crates gives an advice on changing a `format!` without any argument with `.to_string()`. This PR applies the advice.


jyhi:
Oops, did not test locally.

Thoralf-M (approved):

PR #749, opened 2021-11-25
Fix `Cargo.toml`

### LINKED ISSUE #744: Change authors in config file
## Description

Change the authors to `"IOTA Stiftung"` in the config file. 

## Motivation

Should be consistent across projects. Bee & stronghold use `"IOTA Stiftung"` as authors. 

## Open questions (optional)

N/A

Maybe.

jyhi:
This PR tries to improve the quality of `Cargo.toml`s.

Fix #744.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,9 +4,12 @@ version = "1.1.0"
 authors = ["IOTA Stiftung"]
 edition = "2018"
 description = "The official, general-purpose IOTA client library in Rust for interaction with the IOTA network (Tangle)"
+documentation = "https://wiki.iota.org/iota.rs/welcome"
+homepage = "https://www.iota.org/"
 repository = "https://github.com/iotaledger/iota.rs"
 license = "Apache-2.0"
 keywords = ["iota", "tangle", "client"]
+categories = ["cryptography::cryptocurrencies"]
 
 [package.metadata.cargo-udeps.ignore]
 normal = ["async-trait"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,12 +3,10 @@ name = "iota-client"
 version = "1.1.0"
 authors = ["IOTA Stiftung"]
 edition = "2018"
-description = "Client to use IOTA APIs"
-readme = "README.md"
+description = "The official, general-purpose IOTA client library in Rust for interaction with the IOTA network (Tangle)"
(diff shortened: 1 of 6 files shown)


bindings/nodejs/native/Cargo.toml
```diff
@@ -1,12 +1,12 @@
-[workspace]
-
 [package]
-name = "iota-client"
+name = "iota-client-nodejs"
```
Thoralf-M: Here it's the same issue as for wallet.rs `The crate name will be read from the npm_package_name environment variable. If the package name includes a namespace (@namespace/package), the namespace will be removed when matching the crate name (package).`

bindings/wasm/native/Cargo.toml
```diff
@@ -1,16 +1,12 @@
-[workspace]
-
 [package]
-name = "client-wasm"
+name = "iota-client-wasm"
```
Thoralf-M: I think here it's also related to the npm package name, but needs to be verified

bindings/nodejs/native/Cargo.toml
jyhi: So it's `cargo-cp-artifact` having such behavior. iota.rs does not use it during the Node.js / WASM bindings build. Will it be added in the future?

jyhi: I'm fixing this anyways, we don't publish the crates.

Cargo.toml
```diff
@@ -1,21 +1,12 @@
...
-  "Nathan Jaremko <nathan@jaremko.ca>",
-  "Yu-Wei Wu <wusyong9104@gmail.com>"
-]
+authors = ["IOTA Stiftung"]
 edition = "2018"
-description = "Client to use IOTA APIs"
-readme = "README.md"
+description = "The official, general-purpose IOTA client library in Rust for interaction with the IOTA network (Tangle)"
 repository = "https://github.com/iotaledger/iota.rs"
-homepage = "https://www.iota.org"
```
Thoralf-M, line 16: Why do you want to remove this?

jyhi: I think it's some copy-pasta mistakes -- wallet doesn't have one either.

Thoralf-M (approved):

PR #764, opened 2021-12-04
Make get_node public

Thoralf-M:
Can be useful to get a node URL after initialization of the Client


jyhi:
Should this also be exposed in the bindings?

Thoralf-M:
> Should this also be exposed in the bindings?

If nobody requests it, I think we don't need to add it there. I only need it public for the participation/staking stuff in wallet.rs

jyhi (approved):

PR #775, opened 2022-01-01
Allow setting user agent for HTTP requests

rajivshah3:
This PR creates a default user agent for HTTP requests (`iota-client/[version]`) so that we have a user agent instead of an empty string. Library consumers such as `wallet.rs` and Firefly can also set their own if desired.

Still to do: Ask maintainers of `rumqttc` if we can set the user agent of the [initial HTTP request](https://github.com/bytebeamio/rumqtt/blob/4b76fc024dedc48428160bd59a03f160b2107277/rumqttc/src/eventloop.rs#L288) before upgrading to a WebSocket connection

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -212,6 +212,13 @@ impl ClientBuilder {
         self
     }
 
+    /// Set User-Agent header for requests
+    /// Default is "iota-client/[version]"
+    pub fn with_user_agent(mut self, user_agent: &str) -> Self {
+        self.node_manager_builder = self.node_manager_builder.with_user_agent(user_agent.into());
+        self
+    }
+
     /// Selects the type of network to get default nodes for it, only "testnet" is supported at the moment.
     /// Nodes that don't belong to this network are ignored. The &str must match a part or all of the networkId returned
     /// in the nodeinfo from a node. For example, if the networkId is `"private-tangle"`, `"tangle"` can be used.
--- a/src/client.rs
+++ b/src/client.rs
@@ -7,7 +7,7 @@ use crate::{
     builder::{ClientBuilder, NetworkInfo, GET_API_TIMEOUT},
     error::*,
     node::*,
-    node_manager::Node,
+    node_manager::{Node, DEFAULT_USER_AGENT},
 };
 use bee_common::packable::Packable;
 use bee_message::{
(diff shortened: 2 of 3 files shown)


jyhi:
Heh, this was also my idea a few days ago!

What I was thinking is a UA string like:

```
iota-client/x.y.z (library/x.y.z)
```

... where `library` is the HTTP client library used. For example, for `ureq`:

```
iota-client/1.1.1 (ureq/2.2.0)
```

There is perhaps no necessity, but this was what I designed.

rajivshah3:
@lmy441900 good idea! How should we get the version of the HTTP client?

jyhi:
@rajivshah3 Good question; didn't think about this in depth. Unless the library has such interface built into it (AFAIK neither `ureq` nor `reqwest` has one) or use `build.rs` to extract such information from `Cargo.toml`, there don't seem to be a good way to insert such information into the code.

It's not very graceful to use `build.rs` just for inserting this small piece of information into the code. I think unless there _is_ such statistical need, only keeping the version of the library is sufficient. Just forget my proposal.

rajivshah3:
No worries! I was trying to figure that out as well but the only thing I could figure out was something like `build.rs`. The main reason for this PR is that having no user agent may cause some firewalls to consider requests from `iota.rs` as "suspicious" and we want to avoid that if possible. Hopefully a user agent like `iota-client/[version]` is sufficient

jyhi:
> some firewalls

Is this Cloudflare?

From my experience, Cloudflare thinks a client is suspicious when the UA looks like a faked one. In theory the HTTP client library's default should not cause such problem (it's just like `iota-client/x.y.z`, none of them are well-known to a third-party firewall). Maybe the suspicion comes from other factors?

rajivshah3:
> Maybe the suspicion comes from other factors?

Unfortunately, at least in our Cloudflare logs, it's hard to tell. Cloudflare is citing generic reasons like "HTTP requests causing a high request rate to origin"

thibault-martinez:
Hi @rajivshah3 , it seems this probably won't be done before Shimmer hits the IOTA mainnet so we're going to close this PR and open an issue for Shimmer instead. Please re-open if you really you to have this on mainnet.

PR #780, opened 2022-01-10
Fix getIncludedMessage()

Thoralf-M:
Before it failed with `failed to downcast any to function` because the first argument is the transaction id


Thoralf-M:
Python tests are failing because of the devnet node returning 404

jyhi (approved):

Issue #782, opened 2022-01-11
[Task]: Tracking issue for Github Actions workflow refreshment and test improvements


jyhi:
This is a tracking issue for refurbishing our CI automation workflows and gradually improving the coverage of testing / automation tasks they run.

There are several reasons for splitting the whole task:

1. Initial researches have revealed multiple problems in both of the current workflows and the tests we wrote some time ago.
2. There has been a few pending proposal for improving the workflow; this refurbishment should include them too:
    1. #661
        1. #680
    2. #727
3. We want to also learn from other teams (e.g. identity, streams) on how to organize the workflows better.
4. We want to make them as [reusable](https://docs.github.com/en/actions/learn-github-actions/reusing-workflows) as possible.

Directly merging all the changes would cause too many failures, affecting the PR reviewing process. Thus, multiple changes will be PR-ed into the repository one-by-one.

This is a relatively huge damage so it's hard to list every change that would be made - but every PR should be listed here until all tests are enabled and passed.

## Tasks

- [x] Switch the Node.js build system to yarn (#784)
- [x] **PR new workflows, but with failing tests disabled** (#785)
- [x] Fix obsolete assertions in tests of the Node.js bindings (#792)
- [x] Fix Clippy warnings in the Rust core (#793)
- [x] Fix Clippy warnings in the Java bindings (#794)
- [x] Fix Clippy warnings in the Node.js bindings
- [x] Fix Clippy warnings in the Python bindings (#796)
- [x] Fix Clippy warnings in the WebAssembly bindings (#797)
- [x] Fix security auditing problems (#795)

PR #785, opened 2022-01-19
.github/workflows: refresh

jyhi:
This PR refurbishes workflow files, making them more tidy and sound.

Unsuccessful checks are disabled, and will be enabled one by one after the corresponding problems are solved in the source code. See #782.

Fixes:

- missing conditions triggering workflows
- inconsistent copy-pastes across workflows
- comments by adding clear names to jobs and tasks
- caching that did not work across parallel jobs
- builds and releases on outdated Node.js and Python versions
- tools used for testing
- missing coverages by adding more build and test checks
- organization of workflows by using a per-package approach

This PR depends on #784.


.github/actions/setup-rust/action.yml
```diff
@@ -0,0 +1,91 @@
...
+  toolchain:
+    description: Type of toolchain to install.
+    type: string
+    required: false
+    default: stable
+  target:
+    description: Target triple of the toolchain.
+    type: string
+    required: false
+    default:
```
Thoralf-M, line 13: Is this needed if it's just empty?

.github/workflows/python-bindings-publish.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Build Wheels
+        working-directory: bindings/python/native
+        run: |
+          python3 -m pip install -v maturin
+          maturin build --release --manylinux off
+
+      - name: Upload Wheels
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.identifier }}-iota-wallet-py${{ matrix.python }}-wheel
```
Thoralf-M:
```suggestion
          name: ${{ matrix.identifier }}-iota-client-py${{ matrix.python }}-wheel
```

.github/workflows/examine-core.yml
```diff
@@ -0,0 +1,153 @@
...
+    paths:
+      - '.github/workflows/examine-core.yml'
+      - '.github/actions/**'
+      - 'src/**'
+      - 'tests/**'
+      - 'examples/**'
+      - 'Cargo.toml'
+      - 'Cargo.lock'
+  schedule:
+    - cron: '0 1 * * *'
```
Thoralf-M, line 26: With that it will run automatically every day, additionally to the pull requests to the dev branch?

.github/actions/setup-rust/action.yml
jyhi: This is just an explicit empty default, actually redundant.

.github/workflows/python-bindings-publish.yml
jyhi: Oh no.

.github/workflows/examine-core.yml
jyhi: Additionally every job has an `if` statement constraining when they'll be run. For example, none but `audit` will be run every day because all other jobs are marked as `if: ${{ ! github.event.schedule }}`.

This is actually a trick to merge the auditing workflow with the crate-wise workflows. `audit.yml` used to be a separate workflow triggered every day. Now, all crate-wise workflows are triggered every day, but when the workflows are run because of the `schedule` event, only the `audit` job will be run. This reduces redundancy.

.github/workflows/python-bindings-publish.yml
```diff
@@ -5,11 +5,13 @@ on:
     branches:
       - dev
     paths:
-      - .github/workflows/python-bindings-publish.yml
-      - bindings/python/**
-      - src/**
-      - Cargo.lock
-      - Cargo.toml
+      - '.github/workflows/examine-wasm-bindings.yml'
```
Thoralf-M: This should stay python, no?

jyhi: :(

.github/workflows/covector.yml
```diff
@@ -0,0 +1,156 @@
+name: Covector
+on:
+  push:
+    branches:
+      - dev
+    paths-ignore:
+      - 'specs/**'
+      - 'documentation/**'
+  pull_request:
```
rajivshah3: I might be wrong but does this need to run on PRs? The `status` job just seems to [provide outputs](https://github.com/jbolda/covector/tree/covector-v0/packages/action#outputs) for `version-or-publish` which should only run on push events

jyhi: Hmm.

.github/workflows/python-bindings-publish.yml
```diff
@@ -0,0 +1,59 @@
...
+      - 'Cargo.toml'
+      - 'Cargo.lock'
+
+jobs:
+  wheels:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
rajivshah3:
```suggestion
        os: [ubuntu-18.04, macos-latest, windows-latest]
```

I think we build on Ubuntu 18.04 to avoid issues with glibc (#588)

.github/workflows/covector.yml
```diff
@@ -0,0 +1,156 @@
...
+
+  nodejs-binding-prebuild:
+    name: Build Prebuilt Binaries for Node.js
+    if: ${{ needs.version-or-publish.outputs.successfulPublish == 'true' }}
+    needs: version-or-publish
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
rajivshah3:
```suggestion
        os: [ubuntu-18.04, macos-latest, windows-latest]
```

I think we build on Ubuntu 18.04 to avoid issues with glibc (#576)

.github/workflows/python-bindings-publish.yml
jyhi: Fair enough.

Thoralf-M (approved):

PR #786, opened 2022-01-20
Support native TLS certs

rajivshah3:
- feature: Expose native TLS certs via features
- Use wss for secure WebSocket and load native certificates

See iotaledger/iota-sdk#56, iotaledger/firefly#2111

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -284,6 +284,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "core-foundation"
+version = "0.9.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


rajivshah3:
I had a bit of trouble with the features, so I'm not sure what the best way to do those is 🤔

thibault-martinez:
What's the status on that @rajivshah3 ?

rajivshah3:
@thibault-martinez I haven't had time to work on this in a while unfortunately. When I have some more time I can update this PR and test it

thibault-martinez:
Hi @rajivshah3, it seems this probably won't be done before Shimmer hits the IOTA mainnet so we're going to close this PR and open an issue for Shimmer instead. Please re-open if you really you to have this on mainnet.

PR #791, opened 2022-01-31
Add test for address generation with a mnemonic

Thoralf-M:


tests/addresses.rs
```diff
@@ -40,3 +40,26 @@ async fn public_key_to_address() {
         "atoi1qzt0nhsf38nh6rs4p6zs5knqp6psgha9wsv74uajqgjmwc75ugupx3y7x0r".to_string()
     );
 }
+#[tokio::test]
+async fn mnemonic_address_generation() {
+    let iota = Client::builder().with_offline_mode().finish().await.unwrap();
+
+    let mnemnonic = "acoustic trophy damage hint search taste love bicycle foster cradle brown govern endless depend situate athlete pudding blame question genius transfer van random vast";
```
laumair, line 47: @Thoralf-M I think this should read from `tests/fixtures/test_vectors.json` so that the changes are reflected in every test if we made a change in the `test_vectors.json` file.

Thoralf-M: We can change it, but then it shouldn't be in the Python folder

laumair: @Thoralf-M It is not at the root level on in python. @bingyanglin moved it to the root level.

Thoralf-M: But there are also changes in https://github.com/iotaledger/iota.rs/blob/9aef4a5e611087a5caa70dd0efe3e4e235080438/bindings/python/native/tests/fixtures/test_vectors.json, they shouldn't be needed then?

bingyanglin: The [/python/native/tests/fixtures/test_vectors.json](https://github.com/iotaledger/iota.rs/blob/9aef4a5e611087a5caa70dd0efe3e4e235080438/bindings/python/native/tests/fixtures/test_vectors.json) is still needed for other python testings except for the address generation testing, because in different bindings we have different test cases and test vectors, and now the common test case is the address generation, whose test vectors are in the common test file.

bingyanglin: I moved the test vectors which is specific to python binding to the global file, so it can be removed now.

Thoralf-M: But in wallet.rs and not iota.rs?

bingyanglin: Also in iota.rs too. I changed that before and just forgot 😅

thibault-martinez: @Thoralf-M Is this resolved ?

thibault-martinez (approved):

tests/fixtures/test_vectors.json
```diff
@@ -0,0 +1,621 @@
+{
+    "general": {
+        "MNEMNONIC": "acoustic trophy damage hint search taste love bicycle foster cradle brown govern endless depend situate athlete pudding blame question genius transfer van random vast",
+        "MNEMNONIC_ADDRESS": "iota1qpg2xkj66wwgn8p2ggnp7p582gj8g6p79us5hve2tsudzpsr2ap4skprwjg"
+    },
+    "python": {
```
samuel-rufi, line 6: do we really need language specific vectors?

Thoralf-M: @samuel-rufi it's currently used only for the Python tests so there not two files need to be imported and at least I don't plan to adapt the tests in the other languages to the same format
For stardust I would also not continue like this

samuel-rufi: just for better understanding, you plan to remove the test_vectors.json in future completely?

Thoralf-M: It was never added like this there, I only added the addresses https://github.com/iotaledger/iota.rs/blob/develop/tests/fixtures/test_vectors.json

samuel-rufi: okay i see thanks for explaining

samuel-rufi (approved):

tests/addresses.rs
Thoralf-M: Yes

PR #792, opened 2022-01-31
bindings/nodejs: fix minPoWScore assertion on devnet

jyhi:
This PR fixes an obsolete assertion on the devnet about the minimum PoW score.

See also #782.

### DIFF
--- a/.github/workflows/examine-nodejs-bindings.yml
+++ b/.github/workflows/examine-nodejs-bindings.yml
@@ -67,10 +67,10 @@ jobs:
         uses: ./.github/actions/setup-clang
 
       # This step is required for node-gyp to work on Windows with Node.js 16.
-      - name: Set npm config to msvs_version 2017
+      - name: Set npm install -g windows-build-tools
         if: ${{ startsWith(matrix.os, 'windows') }} && ${{ matrix.node == 16 }}
         working-directory: bindings/nodejs/
-        run: npm config set msvs_version 2017
+        run: npm install -g windows-build-tools
 
       - name: Run Yarn Install
         working-directory: bindings/nodejs/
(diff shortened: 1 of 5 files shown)


Thoralf-M (approved):

jyhi:
There turns out to be more problems... can't merge yet.

jyhi:
@Thoralf-M I'm not sure how to change the tests to let them pass.

Thoralf-M:
@lmy441900 I requested more funds from the faucet to the addresses and changed the addresses to one with a higher index, so the transactions will now move funds to the next address (sending to their own address is currently not possible if the remainder would also go to it, this will change with stardust)

Thoralf-M:
But since we run them in parallel we get conflicting transactions, maybe better to disable the transaction tests

jyhi:
@Thoralf-M I have no idea about the failures on Node.js 16 on Windows.

Thoralf-M:
@lmy441900 seems like we need to install Python https://github.com/neon-bindings/neon/issues/844

Thoralf-M:
Btw `Examine the WebAssembly Bindings / Test` fails with a neon error, do these lines need to be changed to `bindings/wasm`?
https://github.com/iotaledger/iota.rs/blob/b25d6687e6fedd38ee44f7c99f448a2b1f4ca9bb/.github/workflows/examine-wasm-bindings.yml#L68
https://github.com/iotaledger/iota.rs/blob/b25d6687e6fedd38ee44f7c99f448a2b1f4ca9bb/.github/workflows/examine-wasm-bindings.yml#L78

Also I'm just seeing we install clang/LLVM, but I think we only need that for the node.js binding and not for wasm

PR #843, opened 2022-03-10
Merge transaction signing code from signers

### LINKED ISSUE #836: [Task]: Merge transaction signing code from signers
### Task description

Merge transaction signing code from menmonic and stronghold signers, only the signature unlock block creation should be different

### Requirements

Create a single function which will be called by both signers

### Acceptance criteria

Signing should work without duplicated code

### Creation checklist

jyhi:
Avoid duplication by reusing the loop going through transaction inputs and outputs. To achieve this without removing it from a `Signer`, `sign_transaction_essence()` is provided as a default implementation of `Signer`. `signature_unlock()` is added to `Signer` for performing the actual signs.

Fix #836.

### DIFF
--- a/src/signing/mod.rs
+++ b/src/signing/mod.rs
@@ -13,10 +13,11 @@ use crate::signing::{
 };
 
 use bee_message::{
-    address::Address,
+    address::{Address, AliasAddress, Ed25519Address, NftAddress},
+    output::Output,
     payload::transaction::{TransactionEssence, TransactionPayload},
     signature::Signature,
-    unlock_block::{UnlockBlock, UnlockBlocks},
+    unlock_block::{AliasUnlockBlock, NftUnlockBlock, ReferenceUnlockBlock, UnlockBlock, UnlockBlocks},
 };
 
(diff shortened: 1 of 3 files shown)


src/signing/mod.rs
```diff
@@ -108,13 +110,76 @@ pub trait Signer {
...
+        metadata: &SignMessageMetadata<'a>,
+    ) -> crate::Result<UnlockBlock>;
     /// Signs transaction essence.
     async fn sign_transaction_essence<'a>(
         &mut self,
         essence: &TransactionEssence,
         inputs: &mut Vec<InputSigningData>,
         metadata: SignMessageMetadata<'a>,
-    ) -> crate::Result<Vec<UnlockBlock>>;
+    ) -> crate::Result<Vec<UnlockBlock>> {
```
Thoralf-M, line 126: Nice, I wasn't aware that one can do default implementations this way, we should do the same for `get_ledger_status()` and `store_mnemonic()`, opened an issue for that https://github.com/iotaledger/iota.rs/issues/844

Thoralf-M (approved):

PR #845, opened 2022-03-10
signing: provide default implementations

### LINKED ISSUE #836: [Task]: Merge transaction signing code from signers

### LINKED ISSUE #844: [Task]: Add default Signer impl for `get_ledger_status()` and `store_mnemonic()`
### Task description

Add default Signer impl for `get_ledger_status()` and `store_mnemonic()`, so we don't have to add it for all Signers where we might don't care about them.
`store_mnemonic()` should by default return an error, that this signer doesn't support storing a mnemonic and `get_ledger_status()` should return 
```Rust
        crate::signing::LedgerStatus {
            connected: false,
            locked: false,
            app: None,
        }
```

### Requirements

Add default impl and remove it from the other signers where it's not required

### Acceptance criteria

Added default impl and removed from the signers where it's not required

### Creation checklist

jyhi:
`Signer::get_ledger_status()` and `Signer::store_mnemonic()` are only for some signers; with default implementations they don't need to be implemented as stub methods any more - they're by default stubs now.

Fix #844.

### DIFF
--- a/src/signing/mod.rs
+++ b/src/signing/mod.rs
@@ -97,9 +97,23 @@ impl Deref for SignerHandle {
 #[async_trait::async_trait]
 pub trait Signer: Send + Sync {
     /// Get the ledger status.
-    async fn get_ledger_status(&self, is_simulator: bool) -> LedgerStatus;
+    ///
+    /// This is only meaningful for the Ledger hardware; other signers don't implement this.
+    async fn get_ledger_status(&self, _: bool) -> LedgerStatus {
+        LedgerStatus {
+            app: None,
+            connected: false,
+            locked: false,
+        }
+    }
+
     /// Initialises a mnemonic.
-    async fn store_mnemonic(&mut self, storage_path: &Path, mnemonic: String) -> crate::Result<()>;
+    ///
+    /// This is only meaningful for the Stronghold signer; other signers don't implement this.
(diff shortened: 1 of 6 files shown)


src/signing/mod.rs
```diff
@@ -97,9 +97,23 @@ impl Deref for SignerHandle {
...
+            locked: false,
+        }
+    }
+
     /// Initialises a mnemonic.
-    async fn store_mnemonic(&mut self, storage_path: &Path, mnemonic: String) -> crate::Result<()>;
+    ///
+    /// This is only meaningful for the Stronghold signer; other signers don't implement this.
+    async fn store_mnemonic(&mut self, _: &Path, _: String) -> crate::Result<()> {
+        Ok(())
```
Thoralf-M: Should we rather return an error, because a user could otherwise expect that it gets stored _somewhere_?

jyhi: Which error can we return, or we crate a new error specifically for this case?

Thoralf-M: We would need a new one

Thoralf-M (approved):

PR #855, opened 2022-03-16
Downgrade zeroize

Thoralf-M:
# Description of change

Downgrade zeroize, because it was requested from the identity team, since they get a conflict with the latest version.


jyhi:
:thinking:

Thoralf-M:
They use https://github.com/RustCrypto/AEADs/blob/e13da9c77aa754abe1a2c91b34e6b22ec1eece8b/chacha20poly1305/Cargo.toml
and there zeroize needs to be `zeroize = { version = ">=1, <1.5", default-features = false }`

jyhi:
Okay, maybe follow with a `cargo update`

jyhi (approved):
Approve because it's already tested.

Issue #798, opened 2022-02-02
[Task]: Implement message translator / actor system in shimmer branch


laumair:
## Description

We have an [implementation](https://github.com/iotaledger/wallet.rs/tree/dev/src/actor) of an actor system (need to rethink about the name) / message passing approach in [wallet-rs](https://github.com/iotaledger/wallet.rs). This needs to be moved to the shimmer branch. 

## Motivation

As we work towards supporting the [shimmer network](https://github.com/iotaledger/iota.rs/commits/shimmer-develop), we should use this approach in iota-rs as well. With that implemented, bindings could be written on top of this implementation. 

## Requirements

- Move actor / message passing approach to [shimmer](https://github.com/iotaledger/iota.rs/tree/shimmer-develop) branch;
- Document this explaining the motivation, usage so that it is easier for bindings to be written using this approach

## Open questions (optional)

- Can the current implementation be improved somehow?
- As similar implementation is also done in [wallet-rs](https://github.com/iotaledger/wallet.rs), could this implementation be moved to an independent crate and then be reused in both somehow? 

## Are you planning to do it yourself in a pull request?

No.

laumair:
Similar ticket for [wallet.rs](https://github.com/iotaledger/wallet.rs/issues/864)

PR #858, opened 2022-03-17
Message interface

pelumi-ajayi:
# Description of change

Implement message interface for iota.rs

## Pending changes
MQTT
Message interface for `get_pow_provider`

Closes  issue #798 

## Type of change

Basic offline tests

## Change checklist

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -6,7 +6,7 @@
 use bee_message::output::TokenId;
 
 use primitive_types::U256;
-
+use serde::ser::{SerializeStruct, Serializer};
 use std::collections::HashMap;
 
 /// Type alias of `Result` in iota-client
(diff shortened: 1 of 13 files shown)


bingyanglin (approved):

src/error.rs
```diff
@@ -218,3 +218,91 @@ impl From<iota_ledger::api::errors::APIError> for Error {
...
+            serializer: S,
+            variant_name: &str,
+        ) -> std::result::Result<S::Ok, S::Error> {
+            let mut state = serializer.serialize_struct("Error", 2)?;
+            state.serialize_field("type", variant_name)?;
+            state.serialize_field("error", &error.to_string())?;
+            state.end()
+        }
+
+        match self {
```
Thoralf-M: Instead of mapping all manually, is there no macro that can do that for us?

pelumi-ajayi: Good point, I was going to use `#[serde(tag = "type", content = "error")` but I ran into errors because some of the variants don't implement serialize, `FromHexError(#[from] hex::FromHexError),`, `CommonError(#[from] bee_common::logger::Error)` and a few others.

https://github.com/iotaledger/iota.rs/blob/shimmer-develop/src/error.rs#L63

Thoralf-M: But for this case we don't care about the errors inside anyways, so maybe we can use the enum discriminant with this approach https://stackoverflow.com/a/32712140 or something similar?

pelumi-ajayi: I think I see what you're saying, we don't need the errors inside but the error message generated by `thiserror`, so maybe the manual `Serialize` implementation don't have to be exhaustive, right?

src/error.rs
```diff
@@ -187,6 +204,7 @@ pub enum Error {
     /// Riker system error during Stronghold initialization
     #[cfg(feature = "stronghold")]
     #[error("Stronghold reported a system error: {0}")]
```
pelumi-ajayi, line 206: I've decided to use `Display` trait for serializing error types that don't implement Serialize, the only slight downside here is that it ignores the string message generated by the `#[error()]` macro which I think is fine.

src/message_interface/response_type.rs
```diff
@@ -0,0 +1,130 @@
...
+    IsAddressValid(bool),
+    /// Generated mnemonic
+    GeneratedMnemonic(String),
+    /// Mnemonic to hex encoded seed
+    MnemonicHexSeed(String),
+    /// An error occurred.
+    Error(Error),
+    /// A panic occurred.
+    Panic(String),
+    /// GenerateMnemonic response.
```
Thoralf-M: Should this be removed?
```suggestion
```

src/message_interface/response_type.rs
```diff
@@ -0,0 +1,130 @@
...
+#[serde(tag = "type", content = "payload")]
+pub enum ResponseType<'a> {
+    /// GenerateAddress response.
+    GeneratedAddresses(Vec<String>),
+    /// Generated message
+    GeneratedMessage(PreparedTransactionData),
+    /// Node
+    Node(Node),
+    // /// Proof of work provider
+    // PoWProvider(ClientMiner),
```
Thoralf-M: The client miner is only required internally so we can set the parameters from the PoW, but for bindings it's useless
```suggestion
```

Thoralf-M (approved):

PR #875, opened 2022-03-26
Prepare for new wiki build system

Dr-Electron:
# Description of change

This is a first draft of the new wiki build system. It should make things a lot easier for devs. We will also provide a v2 of the wiki-cli to make things even easier ;)
I will keep this a draft until it is ready, but now if you want you can already look at what we are creating.
The wiki can be used by just running `yarn`, `yarn start` and navigating to http://localhost:3000/iota-client/welcome

TO-DOs:
- [x] Test
- [x] Add v2 of Wiki CLI

## Type of change

- [ ] Bug fix (a non-breaking change which fixes an issue)
- [ ] Enhancement (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation Fix


documentation/yarn.lock
```diff
@@ -0,0 +1,10369 @@
+# This file is generated by running "yarn install" inside your project.
```
Thoralf-M, line 1: Do we need this file here in this repo? I think we removed it before because it wasn't necessary and we get all the dependabot PRs then

Dr-Electron: Actually I think we can ignore the yarn.lock file. Good point. I will keep that in mind while testing

Dr-Electron: We decided we wanted to make sure that the yarn.lock files stays. You need to make sure to maintain the plugins you use as we now also import them into the wiki

Dr-Electron:
Ok, most of the things should work now :trollface:
You can test the docs locally with `yarn start`
I also added a test build to check for broken links.
Also please inform the Wiki team (on discord) once you plan to merge as we need to update the wiki that day too :)

Thoralf-M (approved):

.github/workflows/test-docs-build.yml
```diff
@@ -0,0 +1,21 @@
+name: Test Docs Build
```
Thoralf-M, line 1: Should it still be `Test`?

Dr-Electron: Why not? It's still just a test. Mainly to detect broken links

Thoralf-M: Don't mind me, I didn't went through the whole file :trollface:

thibault-martinez (approved):

Issue #830, opened 2022-03-01
[Task]: WASM/Update NetworkInfo in the MessageBuilder before sending a message


Thoralf-M:
## Description

Update NetworkInfo in the MessageBuilder before sending a message
I think it's enough if we update the NetworkInfo inside of `get_network_info()` just always with the wasm feature

## Motivation

For the WASM binding we don't have the node syncing process, which updates it every 60 seconds and in the future if there will be a change via milestones a Client could construct invalid messages then, if we don't update it every time before sending a message

## Requirements

1. Inside of the message builder get the node info and update the NetworkInfo of the client (behind a wasm feature)

## Are you planning to do it yourself in a pull request?

Maybe.

PR #879, opened 2022-03-29
Update networkinfo with wasm feature every time

Thoralf-M:
# Description of change

For the WASM binding we don't have the node syncing process, which updates it every 60 seconds and in the future if there will be a change via milestones a Client could construct invalid messages then, if we don't update it every time before sending a message

Fixes https://github.com/iotaledger/iota.rs/issues/830

## Type of change

## Change checklist

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -279,7 +279,7 @@ impl Client {
         // For WASM we don't have the node syncing process, which updates the network_info every 60 seconds, but the PoW
         // difficulty or the byte cost could change via a milestone, so we request the nodeinfo every time, so we don't
         // create invalid transactions/messages
-        if not_synced || wasm {
+        if not_synced || cfg!(feature = "wasm") {
             let info = self.get_info().await?.nodeinfo;
             let network_id = hash_network(&info.protocol.network_name).ok();
             {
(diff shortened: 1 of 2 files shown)


src/client.rs
```diff
@@ -272,7 +272,14 @@ impl Client {
     pub async fn get_network_info(&self) -> Result<NetworkInfo> {
         let not_synced = self.network_info.read().map_or(true, |info| info.network_id.is_none());
 
-        if not_synced {
+        #[cfg(not(feature = "wasm"))]
+        let wasm = false;
+        #[cfg(feature = "wasm")]
+        let wasm = true;
```
lmy441900, line 278: Would this be better?

```suggestion
        let wasm = if cfg!(feature = "wasm") { true } else { false };
```

src/client.rs
```diff
@@ -272,7 +272,14 @@ impl Client {
...
+        // For WASM we don't have the node syncing process, which updates the network_info every 60 seconds, but the PoW
+        // difficulty or the byte cost could change via a milestone, so we request the nodeinfo every time, so we don't
+        // create invalid transactions/messages
+        if not_synced || wasm {
```
jyhi: Would this be better?

```suggestion
        if not_synced || cfg!(feature = "wasm") {
```

Thoralf-M: Yes, thanks! Forgot about that option because in most places we can't use it like this

pelumi-ajayi (approved):

PR #883, opened 2022-03-31
stronghold,signing,db: add StrongholdAdapter

### LINKED ISSUE #834: [Task]: Extend Stronghold Signer functionality
### Task description

It should be able to use the stronghold file for backups, so we need functions to store data inside and also read it again.
For the password to improve security we also want to add the option for an auto clearing from the memory after x seconds, but also need functions to set it again later and also to manually clear it.

### Requirements

1. Add a function to store bytes at a location
2. Add a function to get bytes that are stored at a location
3. Add auto clearing of the password (spawn a thread which does this, similar to https://github.com/iotaledger/wallet.rs/blob/213140abefbe6aec0d0d15d5a151b408e75fcc1e/src/stronghold.rs#L209), if the password is cleared we should also emit an event for that, so applications can know that
4. Add an optional field to set the timeout during initialisation, if nothing is provided, it should be disabled
5. Add a function to set the password (needed to unlock it, if the password got cleared before)
6. Add a function to clear the password

### Acceptance criteria

We want to use it in wallet.rs for backups and improve the security for applications like Firefly

### Creation checklist

jyhi:
StrongholdAdapter acts as an all-in-one Stronghold integration for iota.rs that can both act as a key-value database provider and a transaction signer. StrongholdSigner and StrongholdDatabaseProvider are provided as type aliases for naming consistency.

Add db module for the definition of DatabaseProvider trait.

Fix #834.

### DIFF
--- a/src/signing/mod.rs
+++ b/src/signing/mod.rs
@@ -4,9 +4,9 @@
 //! Signing module to allow using different signer types for address generation and transaction essence signing
 
 #[cfg(feature = "ledger")]
-use crate::signing::ledger::LedgerSigner;
+use self::ledger::LedgerSigner;
 #[cfg(feature = "stronghold")]
-use crate::signing::stronghold::StrongholdSigner;
+use self::stronghold::StrongholdSigner;
 use crate::signing::{
     mnemonic::MnemonicSigner,
     types::{InputSigningData, SignerTypeDto},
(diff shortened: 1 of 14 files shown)


src/signing/mod.rs
```diff
@@ -74,10 +76,11 @@ impl SignerHandle {
 
         Ok(match signer_type {
             #[cfg(feature = "stronghold")]
-            SignerTypeDto::Stronghold(stronghold_dto) => StrongholdSigner::try_new_signer_handle(
-                &stronghold_dto.password,
-                Path::new(&stronghold_dto.snapshot_path),
-            )?,
+            SignerTypeDto::Stronghold(stronghold_dto) => StrongholdSigner::builder()
+                .password(&stronghold_dto.password)
```
Thoralf-M: Please make the password in the dto optional and only set it here if it's `Some`

Thoralf-M (approved):

PR #908, opened 2022-04-12
signing: refactor trait Signer, struct *Signer

jyhi:
The Signer trait now gets a few more general methods for every signer implementation to implement: type, init, sync, set_password, clear_password, gen_addrs (prev. generate_addresses), and unlock (prev. signature_unlock).

SignerHandle is removed because now Signer implementations can supply their typing information via signer_type(). Related code are refactored too.

MnemonicSigner and LedgerSigner w/ their usages around the repo have been refactored, although todo methods are still left todo.

StrongholdAdapter also creates a backup every time a snapshot is flushed to the disk by copying the just-written snapshot to a new one with a UNIX timestamp in the filename.

This PR is an extension of #883, and a part of #834.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ regex = { version = "1.5.4", default-features = false, features = ["std", "unico
 serde = { version = "1.0.130", default-features = false, features = ["derive"] }
 serde_json = { version = "1.0.68", default-features = false }
 thiserror = { version = "1.0.29", default-features = false }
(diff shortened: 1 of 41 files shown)


jyhi:
Don't merge yet, I'll try addressing some CI errors first.

jyhi:
There are still design problems to be discussed - don't merge.

jyhi:
I'll reopen a PR for the discussed changes on the design.

Issue #801, opened 2022-02-10
[Task]: Add high level function to get inputs from a transaction/message


Thoralf-M:
## Description
Add high level function to get inputs from a transaction/message; __NOT__ just their ID, but the full output with an amount and address.

The typical data flow should look like...
1. Get transaction by transaction or message ID
2. Get outputs from the input IDs
3. Return the outputs

## Motivation
Many people want to know the inputs for different reasons and it should be easy to get them.

## Requirements
- [ ] Implement high-level function for retrieving inputs via transaction ID
- [ ] Implement high-level function for retrieving inputs via message ID

## Open questions (optional)
- Do we want to accept the message or transaction ID? (__answer__: transaction ID)

## Are you planning to do it yourself in a pull request?
No.

maxwellmattryan:
I think the transaction ID would be better since it is closer in the domain to UTXOs / value transfer. Although I don't see why we couldn't have a `get_inputs_from_message` function that would use the underlying `get_inputs_from_transaction` function after some additional checks / logic. What do you think?

Thoralf-M:
Sure, we can also add both

samuel-rufi:
Implemented by https://github.com/iotaledger/iota.rs/pull/909

PR #909, opened 2022-04-12
Introduce `GetTransactionInputsBuilder` to fetch inputs from a given transaction id

samuel-rufi:
# Description of change

Fixes https://github.com/iotaledger/iota.rs/issues/801

Be sure to reference any related issues by adding `fixes issue #`.

## Type of change

Describe the tests that you ran to verify your changes.

Make sure to provide instructions for the maintainer as well as any relevant configurations.

## Change checklist

- [X] I have followed the contribution guidelines for this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have checked that new and existing unit tests pass locally with my changes

### DIFF
--- a/examples/inputs_from_transaction_id.rs
+++ b/examples/inputs_from_transaction_id.rs
@@ -1,0 +1,31 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! cargo run --example inputs_from_transaction_id --release
+
+use iota_client::{
+    bee_message::payload::transaction::TransactionId, node_api::high_level::GetTransactionInputsBuilder, Client, Result,
+};
+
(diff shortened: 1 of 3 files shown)


Thoralf-M:
Currently we also also have all high level methods available from the Client directly https://github.com/iotaledger/iota.rs/blob/93610fd2576ab4799672a614eb1e28c752a4426d/src/client.rs#L591, I think this should also be added there (maybe all high level methods should be moved to another file that has `impl Client {}`, but that's a more general question)
I also think that the builder is a little bit too much for this case, because the user only provides the transaction id and nothing else

PR #910, opened 2022-04-13
Semantic validation after transaction signing

thibault-martinez:
## Change checklist


examples/offline_signing/3_send_message.rs
```diff
@@ -22,6 +27,25 @@ async fn main() -> Result<()> {
 
     let signed_transaction_payload = read_signed_transaction_from_file(SIGNED_TRANSACTION_FILE_NAME)?;
 
+    // TODO @thibault-martinez: I don't like that we have to refetch the prepared transaction. Will revisit later.
```
Thoralf-M, line 30: Yeah that's annoying, maybe store the input_signing_data_entries separated and the build the prepared PreparedTransactionData from the signed transaction essence with it together

src/api/message_builder/transaction.rs
```diff
@@ -137,9 +139,55 @@ pub async fn sign_transaction(
...
+        .iter()
+        .map(|i| Ok(Output::try_from(&i.output_response.output)?))
+        .collect::<Result<Vec<Output>>>()?;
+    let inputs = prepared_transaction_data
+        .input_signing_data_entries
+        .iter()
+        .zip(outputs.iter())
+        .map(|(i, o)| {
+            Ok((
+                // PANIC: at this point, OutputIds have been validated many times already.
```
Thoralf-M: If it gets read from the file like the offline example, then it might not be validated before and maybe there are also other use cases, so not sure if we should assume this

src/client.rs
```diff
@@ -918,6 +918,23 @@ impl Client {
         Ok((message_id, msg))
     }
 
+    /// Rerturns checked local time and milestone index.
```
Thoralf-M:
```suggestion
    /// Returns checked local time and milestone index.
```

src/error.rs
```diff
@@ -232,6 +232,12 @@ pub enum Error {
     #[cfg(feature = "stronghold")]
     #[error("no snapshot path has been supplied")]
     StrongholdSnapshotPathMissing,
+    /// The semantic validation of a transaction failed.
+    #[error("the semantic validation of a transaction failed")]
+    TransactionSemantic(ConflictReason),
+    /// Local time doesn't match the time of the latest timestamp
+    #[error("Local time {0} doesn't match the time of the latest timestamp: {1}")]
```
Thoralf-M: I know I wrote it in wallet.rs like this, but maybe we should use `ms timestamp` or even `milestone timestamp` so it's clearer?

examples/offline_signing/3_send_message.rs
thibault-martinez: Do you mind if I let it like that temporarily ? I'd like to solve it another way in a future PR.

Thoralf-M (approved):

PR #922, opened 2022-04-21
Change Signer to SecretManager

### LINKED ISSUE #834: [Task]: Extend Stronghold Signer functionality

jyhi:
`Signer` is now `SecretManager`, under `crate::secret`. Names and module paths across the Rust code are changed.

`enum SecretManagerType` replaces `struct SignerHandle` as a design improvement. `SecretManager` (previously `Signer`) has slight improvements as well. Examples and tests are changed and improved.

`MnemonicSecretManager` and `LedgerSecretManager` have their code refined and constructor changed.

This PR as a successor of #908 fixes #834.


src/secret/mnemonic.rs
```diff
@@ -13,96 +16,62 @@ use crypto::{
...
-        Ok(SignerHandle::new(
-            SignerType::Mnemonic,
-            Box::new(Self(Seed::from_bytes(&hex::decode(seed)?))),
-        ))
-    }
-}
+/// Secret manager that uses only a mnemonic.
+///
+/// Computation are done in-memory. A mnemonic needs to be supplied upon the creation of [`MnemonicSecretManager`].
+pub struct MnemonicSecretManager(Seed);
```
Thoralf-M, line 25:
```suggestion
#[derive(Clone)]
pub struct MnemonicSecretManager(Seed);
```

jyhi: `Seed` is not `Clone`

Thoralf-M: Oh right, we only had clone for `SignerHandle` because it had an Arc, so all good how it's now

src/message_interface/client_method.rs
```diff
@@ -25,15 +25,17 @@ use crate::{
 pub enum ClientMethod {
     /// Generate a addresses.
     GenerateAddresses {
-        /// Create singer from json
-        signer: String,
+        /// Create secret manager from json; alias "signer" for compatibility
+        #[serde(alias = "signer")]
+        secmngr: String,
```
thibault-martinez: I wouldn't abbreviate for field names, I'd use the complete `secret_manager`.

src/message_interface/client_method.rs
```diff
@@ -25,15 +25,17 @@ use crate::{
...
         /// Addresses generation options
         options: GenerateAddressesOptions,
     },
     /// Generate client message
     GenerateMessage {
-        /// Signer
-        signer: Option<String>,
+        /// Secret manager; alias "signer" for compatibility
+        #[serde(alias = "signer")]
+        secmngr: Option<String>,
```
thibault-martinez: Same

src/message_interface/client_method.rs
```diff
@@ -219,8 +221,9 @@ pub enum ClientMethod {
     /// Function to consolidate all funds from a range of addresses to the address with the lowest index in that range
     /// Returns the address to which the funds got consolidated, if any were available
     ConsolidateFunds {
-        /// Signer
-        signer: String,
+        /// Secret manager; alias "signer" for compatibility
+        #[serde(alias = "signer")]
+        secmngr: String,
```
thibault-martinez: Same

src/api/message_builder/input_selection/automatic.rs
```diff
@@ -209,7 +208,9 @@ async fn get_inputs_for_sender_and_issuer(
     for address in required_ed25519_addresses {
         if let Address::Ed25519(address) = address {
             let (address_index, internal) = search_address(
-                message_builder.signer.ok_or(Error::MissingParameter("Signer"))?,
+                message_builder
+                    .secret_manager
+                    .ok_or(Error::MissingParameter("Signer"))?,
```
Thoralf-M: `secret manager`

src/message_interface/client_method.rs
```diff
@@ -25,15 +25,17 @@ use crate::{
 pub enum ClientMethod {
     /// Generate a addresses.
     GenerateAddresses {
-        /// Create singer from json
-        signer: String,
+        /// Create secret manager from json; alias "signer" for compatibility
+        #[serde(alias = "signer")]
```
Thoralf-M: I would remove the alias and update the comment now, no need to keep compatibility there

src/secret/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+use self::ledger::LedgerSecretManager;
+#[cfg(feature = "stronghold")]
+use self::stronghold::StrongholdSecretManager;
+use self::{mnemonic::MnemonicSecretManager, types::SecretManagerDto};
+use crate::secret::types::InputSigningData;
+
+/// The secret manager interface.
+#[async_trait]
+pub trait SecretManage: Send + Sync {
+    /// Generates an address.
```
Thoralf-M: Or multiple addresses

Thoralf-M: Or multiple addresses

src/secret/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+        account_index: u32,
+        address_indexes: Range<u32>,
+        internal: bool,
+        metadata: GenerateAddressMetadata,
+    ) -> crate::Result<Vec<Address>>;
+
+    /// Sign on `essence`, unlock `input` by returning an [UnlockBlock].
+    ///
+    /// Compared with `sign_transaction_essence()`, this signs on a single input. It's called by
+    /// `sign_transaction_essence()` on every input it receives.
```
Thoralf-M: > on every input it receives.

Remove this part or rewrite, because it's only called if the input requires an Ed25519 signature and if there is no signature for the address already

Thoralf-M: > on every input it receives.

Remove this part or rewrite, because it's only called if the input requires an Ed25519 signature and if there is no signature for the address already

Thoralf-M (approved):

thibault-martinez (approved):

PR #923, opened 2022-04-21
Refactor input selection

Thoralf-M:
# Description of change

Removed hardcoded address for remainder storage deposit calculation and added always all native tokens from inputs
Refactored the way the inputs are selected to make it clearer
Enabled native token burning

## Type of change

- Enhancement

With the examples, in the foundry example we now also burn native tokens and when `.with_burning_allowed(true)` is set to false it still correctly returns an error and doesn't burn the native tokens

## Change checklist


pelumi-ajayi (approved):

Cargo.toml
```diff
@@ -15,15 +15,15 @@ categories = [ "cryptography::cryptocurrencies" ]
...
 [dependencies]
-bee-rest-api = { git = "https://github.com/iotaledger/bee", rev = "e9ef4d6725a9b63756107618c60b75cf1aca56d1", default-features = false }
-bee-message = { git = "https://github.com/iotaledger/bee", rev = "e9ef4d6725a9b63756107618c60b75cf1aca56d1", default-features = false, features = [ "serde1", "dto" ] }
-bee-pow = { git = "https://github.com/iotaledger/bee", rev = "e9ef4d6725a9b63756107618c60b75cf1aca56d1", default-features = false }
+bee-rest-api = { git = "https://github.com/iotaledger/bee", rev = "4b5b74e5b7aedd0f52313af8814673d8a1f571bd", default-features = false }
+bee-message = { git = "https://github.com/iotaledger/bee", rev = "4b5b74e5b7aedd0f52313af8814673d8a1f571bd", default-features = false, features = [ "serde", "dto" ] }
+bee-pow = { git = "https://github.com/iotaledger/bee", rev = "4b5b74e5b7aedd0f52313af8814673d8a1f571bd", default-features = false }
 # bee-rest-api = { path = "../bee/bee-api/bee-rest-api",  default-features = false }
-# bee-message = { path = "../bee/bee-message",  default-features = false, features = [ "serde1" ] }
+# bee-message = { path = "../bee/bee-message",  default-features = false, features = [ "serde" ] }
```
samuel-rufi: why is this dependency not used? can this be removed?

Cargo.toml
```diff
@@ -15,15 +15,15 @@ categories = [ "cryptography::cryptocurrencies" ]
...
 # bee-pow = { path = "../bee/bee-pow",  default-features = false }
```
samuel-rufi: same for this one

src/api/message_builder/input_selection/output_data.rs
```diff
@@ -5,127 +5,28 @@ use std::collections::{hash_map::Entry, HashMap};
...
+    Result,
 };
 
-// Get a remainder with amount and native tokens if necessary, if no remainder_address is provided it will be selected
-// from the inputs, also validates the amounts
-pub(crate) async fn get_remainder(
-    inputs: &[Output],
+// Calculate required accumulated amounts from the outputs, considers also minted and melted native tokens
+pub(crate) async fn get_accumulated_output_amounts(
+    input_outputs: &[Output],
```
samuel-rufi: maybe call it `inputs` simply if that would work?

src/api/message_builder/input_selection/remainder.rs
```diff
@@ -0,0 +1,166 @@
...
+        // get address from an input, by default we only allow ed25519 addresses as remainder, because then we're
+        // sure that the sender can access it
+        None => {
+            let mut address = None;
+            for input in inputs {
+                if let Some(unlock_conditions) = input.unlock_conditions() {
+                    for unlock_condition in unlock_conditions.iter() {
+                        if let UnlockCondition::Address(address_unlock_condition) = unlock_condition {
+                            address.replace(address_unlock_condition.address());
+                            break;
```
samuel-rufi: this only breaks the inner loop, but you actually want to break from both loops right?

src/api/message_builder/input_selection/remainder.rs
```diff
@@ -0,0 +1,166 @@
...
+            return Err(Error::NoBalanceForNativeTokenRemainder);
+        }
+    }
+
+    Ok(remainder_output)
+}
+
+// Return provided remainder address or get an Ed25519 address from the inputs as remainder address
+// We don't want to use nft or alias addresses as remainder address, because we might can't control them later
+pub(crate) fn get_remainder_address(remainder_address: Option<Address>, inputs: &[Output]) -> Result<Address> {
```
samuel-rufi: if `remainder_address` is `Some` then this function actually doens't need to be called right?
So maybe call it only in the `None` case and change the signature to

`pub(crate) fn get_remainder_address(inputs: &[Output]) -> Result<Address> { ... }`

samuel-rufi (changes requested):

Cargo.toml
Thoralf-M: I left it there because I often test new changes with bee locally, but now not so often anymore and if you guys don't use it we can also remove it

src/api/message_builder/input_selection/remainder.rs
Thoralf-M: Good catch!

Cargo.toml
samuel-rufi: okay if you don't need it so much anymore then i guess it's better to remove it

thibault-martinez (approved):
I see a lot of places where we can simplify native tokens usage with the new builder, I'll do that when this is merged.

PR #977, opened 2022-05-04
Add OutputMetadata type

### LINKED ISSUE #921: Separate output metadata from OutputResponse
## Description

Separate output metadata from OutputResponse so we have the metadata next to an actual Output and not an OutputDto

## Motivation

Currently we use OutputResponse a lot and then always need to handle errors from the output conversion if we do `Output::try_from(&o.output_response.output)?` which is annoying

## Requirements

1. Separate Output Metadata
2. Have a new `OutputWithMetadata` struct with the metadata and an Output (name can also be different)
3. Update code to use the new struct so we don't have to handle the errors everywhere

No.

thibault-martinez:
Fixes #921 

I will do another PR to remove a lot of clones that are currently required because of function signatures.
There is also a repetition when creating an OutputMetadata that I'll fix later.

## Change checklist

### DIFF
--- a/src/api/message_builder/input_selection/automatic.rs
+++ b/src/api/message_builder/input_selection/automatic.rs
@@ -3,13 +3,9 @@
 
 //! Automatic input selection for transactions
 
-use std::str::FromStr;
-
 use bee_message::{
     address::Address,
     output::{feature_block::FeatureBlock, ByteCostConfig, Output},
-    payload::transaction::TransactionId,
-    MessageId,
 };
 use crypto::keys::slip10::Chain;
 
@@ -113,21 +109,7 @@ pub(crate) async fn get_inputs(
                 for output_response in address_outputs {
                     available_inputs.push(InputSigningData {
                         output: Output::try_from(&output_response.output)?,
-                        output_metadata: OutputMetadata {
-                            message_id: MessageId::from_str(&output_response.message_id)?,
-                            transaction_id: TransactionId::from_str(&output_response.transaction_id)?,
-                            output_index: output_response.output_index,
-                            is_spent: output_response.is_spent,
-                            milestone_index_spent: output_response.milestone_index_spent,
-                            milestone_timestamp_spent: output_response.milestone_timestamp_spent,
-                            transaction_id_spent: output_response
-                                .transaction_id_spent
-                                .map(|s| TransactionId::from_str(&s))
-                                .transpose()?,
-                            milestone_index_booked: output_response.milestone_index_booked,
-                            milestone_timestamp_booked: output_response.milestone_timestamp_booked,
-                            ledger_index: output_response.ledger_index,
-                        },
+                        output_metadata: OutputMetadata::try_from(&output_response)?,
                         chain: Some(Chain::from_u32_hardened(vec![
                             HD_WALLET_TYPE,
                             message_builder.coin_type,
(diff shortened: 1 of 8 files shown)


src/secret/types.rs
```diff
@@ -88,11 +90,38 @@ pub struct LedgerStatus {
     pub(crate) app: Option<LedgerApp>,
 }
 
+///
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct OutputMetadata {
+    ///
```
thibault-martinez: I'll write docs in a bit.

src/api/message_builder/input_selection/automatic.rs
```diff
@@ -108,7 +112,22 @@ pub(crate) async fn get_inputs(
...
+                            milestone_index_spent: output_response.milestone_index_spent,
+                            milestone_timestamp_spent: output_response.milestone_timestamp_spent,
+                            transaction_id_spent: output_response
+                                .transaction_id_spent
+                                .map(|s| TransactionId::from_str(&s))
+                                .transpose()?,
+                            milestone_index_booked: output_response.milestone_index_booked,
+                            milestone_timestamp_booked: output_response.milestone_timestamp_booked,
+                            ledger_index: output_response.ledger_index,
+                        },
```
Thoralf-M: Shouldn't we just impl TryFrom<&OutputResponse> for OutputMetadata{}?

Thoralf-M: Ah probably what you meant with `There is also a repetition when creating an OutputMetadata that I'll fix later.`, do you also want to do that in another PR?

thibault-martinez: Yeah, I'll do it now.

src/secret/types.rs
```diff
@@ -88,11 +93,61 @@ pub struct LedgerStatus {
     pub(crate) app: Option<LedgerApp>,
 }
 
+///
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct OutputMetadata {
```
Thoralf-M, line 98: Should have `#[serde(rename_all = "camelCase")]` or each field renamed like that

src/secret/types.rs
```diff
@@ -88,11 +93,61 @@ pub struct LedgerStatus {
...
+
 /// Data for transaction inputs for signing and ordering of unlock blocks
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct InputSigningData {
-    /// The output response
-    pub output_response: OutputResponse,
+    /// The output
+    pub output: Output,
+    /// The output metadata
+    pub output_metadata: OutputMetadata,
```
Thoralf-M, line 161:
```suggestion
    #[serde(rename = "outputMetadata")]
    pub output_metadata: OutputMetadata,
```

src/secret/types.rs
thibault-martinez: Yeah I asked on Slack if we could nest all these into a `metadata` field because otherwise we have to redo this all over again :/

thibault-martinez: Will do in the meantime

src/secret/types.rs
thibault-martinez: Why do we need camelCase here ?

Thoralf-M (dismissed):

src/secret/types.rs
```diff
@@ -88,31 +93,93 @@ pub struct LedgerStatus {
     pub(crate) app: Option<LedgerApp>,
 }
 
+///
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
Thoralf-M: Please also add `PartialEq, Eq`

Thoralf-M (approved):

jyhi (approved):

PR #990, opened 2022-05-08
Update and improve tests

samuel-rufi:


pelumi-ajayi (dismissed):

tests/node_api.rs
```diff
@@ -2,232 +2,206 @@
...
+    bech32_to_hex,
     node_api::indexer::query_parameters::QueryParameter,
+    request_funds_from_faucet,
     secret::{mnemonic::MnemonicSecretManager, SecretManager},
+    Client,
 };
 
-const DEFAULT_NODE_URL: &str = "http://localhost:14265";
+const DEFAULT_DEVNET_NODE_URL: &str = "http://localhost:14265";
+const DEFAULT_DEVNET_FAUCET_URL: &str = "http://localhost:14265";
```
Thoralf-M, line 19: Maybe also add a default const for the bech32 HRP and use that for address generation and also use it then for `rms1qzt0nhsf38nh6rs4p6zs5knqp6psgha9wsv74uajqgjmwc75ugupx4aaacx` (needs to be regenerated with the correct hrp)

tests/node_api.rs
```diff
@@ -2,293 +2,265 @@
...
         .await
         .unwrap();
-    println!("{:#?}", r);
+
+    let transaction_id = match message.payload() {
+        Some(Payload::Transaction(t)) => t.id(),
+        _ => unreachable!(),
+    };
+
+    tokio::time::sleep(std::time::Duration::from_secs(20)).await;
```
Thoralf-M: Here I would rather use `let _ = client.retry_until_included(&message.id(), None, None).await?;`, because we know the message, only for the faucet request we don't know it

samuel-rufi: Fixed it

Thoralf-M (approved):

pelumi-ajayi (approved):

PR #991, opened 2022-05-09
Update workflows branches

thibault-martinez:
## Change checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works


.github/workflows/examine-java-bindings.yml
```diff
@@ -15,9 +14,8 @@ on:
       - 'Cargo.lock'
   pull_request:
     branches:
-      # disabled until Java binding is updated
-      # - develop
-      - dev
+      - develop
```
Thoralf-M, line 17: This will fail, since the Java binding wasn't updated yet

.github/workflows/examine-wasm-bindings.yml
```diff
@@ -15,9 +14,8 @@ on:
       - 'Cargo.lock'
   pull_request:
     branches:
-      # disabled until wasm binding is updated
-      # - develop
-      - dev
+      - develop
```
Thoralf-M: Same here, wasm binding was also not updated

.github/workflows/python-bindings-publish.yml
```diff
@@ -3,9 +3,8 @@ name: Build and Upload Wheels for the Python Bindings
 on:
   push:
     branches:
-      # disabled until python binding is updated
-      # - develop
-      - dev
+      - develop
```
Thoralf-M, line 6: Not sure if this should be done like this now because it could interfere with the production builds? @bingyanglin do you know if it would be fine?

thibault-martinez: It won't interfere with production unless develop is merged into production, and when that happens, CI should pass anyway.

bingyanglin: It should be fine.

PR #993, opened 2022-05-10
API spring cleaning

thibault-martinez:
## Change checklist


documentation/docs/node_api_core.md
```diff
@@ -0,0 +1,509 @@
+# `Client::get_health`
```
Thoralf-M, line 1: Why not just link/include the Rust docs for this kind of stuff? Don't really see added value in this

thibault-martinez: Yeah I was kind of thinking the same which is why I didn't actually fill the request/response yet. I'd still like to have this page to give the example commands, so I'll probably end up reducing the content a lot and link to the rust docs. I'll rework this doc page when I'm done with all the example if you're fine with it.

examples/node_api_core/15_get_treasury.rs
```diff
@@ -0,0 +1,28 @@
...
+        .unwrap_or_else(|| "http://localhost:14265".to_string());
+    // Creates a client instance with that node.
+    let client = Client::builder()
+        .with_node(&node)?
+        .with_node_sync_disabled()
+        .finish()
+        .await?;
+
+    // Sends the request.
+    let treasury = client.get_treasury().await?;
```
Thoralf-M, line 22: I think it's not really clear what the treasury is, maybe here or/and in the comment for `get_treasury()` we should add, that `The treasury contains all tokens from the legacy network that have not yet been migrated` or something similar

thibault-martinez: Will do

src/node_api/core/routes.rs
```diff
@@ -264,94 +217,166 @@ impl Client {
...
+        self.node_manager.get_request(path, None, self.get_timeout()).await
+    }
+
+    /// Gets the milestone by the given milestone index.
+    /// GET /api/v2/milestones/{index}
+    pub async fn get_milestone_by_index(&self, index: u32) -> Result<MilestoneResponse> {
+        let path = &format!("api/v2/milestones/by-index/{}", index);
+
+        self.node_manager.get_request(path, None, self.get_timeout()).await
+    }
```
Thoralf-M, line 359: Shouldn't the order be changed so it's first the `get_milestone...` routes and then `get_utxo_changes...`?

thibault-martinez: I'm following the order of the TIP

src/node_manager/http_client.rs
```diff
@@ -53,22 +24,21 @@ impl Response {
         self.0.json().await.map_err(Into::into)
     }
 
+    // TODO is this really needed ?
```
Thoralf-M, line 27: I think it was required for quorum because  the responses didn't had `PartialEq, Eq` available, so we couldn't use them to compare the results, but should be possible now
The other places where it's used are now gone or also can be replaced with `.json()` I think, but needs to be checked

thibault-martinez: Yeah I'll investigate this next 👍🏻

src/node_api/core/routes.rs
```diff
@@ -301,6 +301,7 @@ impl Client {
     }
 
     /// Gets the current treasury output.
+    /// The treasury outputs contains all tokens from the legacy network that have not yet been migrated.
```
Thoralf-M: Shouldn't it be `The treasury output` without `s`?

thibault-martinez: Good catch

Cargo.toml
```diff
@@ -72,15 +68,61 @@ bee-ternary = { version = "0.5.2", default-features = false }
...
-tls = [ "reqwest/rustls-tls", "ureq/tls" ]
+tls = [ "reqwest/rustls-tls" ]
 stronghold = [ "iota_stronghold", "riker", "crypto05" ]
 message_interface = [ "backtrace" ]
 
+# Node API core examples
+
+[[example]]
+name = "node_api_core_get_info"
+path = "examples/node_api_core/02_get_info.rs"
```
samuel-rufi, line 82: should it be renamed to `01_get_info.rs` ?

Thoralf-M (dismissed):

src/message_interface/client_method.rs
```diff
@@ -157,24 +157,24 @@ pub enum ClientMethod {
...
     },
     /// Get the milestone by the given index.
-    GetMilestoneByMilestoneIndex {
+    GetMilestoneByIndex {
         /// Milestone Index
         index: u32,
     },
     /// Get the UTXO changes by the given milestone id.
-    GetUtxoChangesByMilestoneId {
+    GetUtxoChangesById {
```
samuel-rufi, line 171: `GetUtxoChangesById` doesn't tell which id;
maybe better to keep it like `GetUtxoChangesByMilestoneId` ?

samuel-rufi (changes requested):

Cargo.toml
thibault-martinez: No, 01 is get_health but I haven't done it yet

src/message_interface/client_method.rs
thibault-martinez: Idk, it's kind of obvious we're speaking of a milestone. We're putting a suffix to disambiguate id/index but at the end of the day we shouldn't have to put the parameter type name in the method name. Otherwise why not do `get_message_by_message_id` ? Because it's obvious

Thoralf-M (approved):

src/message_interface/client_method.rs
samuel-rufi: good point, more or less obvious

samuel-rufi (approved):

Issue #988, opened 2022-05-07
[Task]: Add .zeroize() to mnemonic and stronghold password


Thoralf-M:
### Task description

Add .zeroize() to mnemonic and stronghold password in store_mnemonic() and also to the other methods in the message_interface (also remove the MessageType in the Response for that, like we did for wallet.rs here https://github.com/iotaledger/wallet.rs/pull/1019/commits/530d3e440688fd5a1755328b0fccce116db02a22)

### Requirements

Stronghold password and mnemonic is zeroized when possible

### Acceptance criteria

N/A

### Creation checklist

- [ ] I have assigned this task to the correct people
- [X] I have added the most appropriate labels
- [X] I have linked the correct milestone and/or project

jyhi:
Fixed by #995.

PR #995, opened 2022-05-10
Stronghold Integration and Message Interface Improvements

### LINKED ISSUE #992: Add a method to stronghold which returns if the key is available
## Description

Method should return if `key: Arc<Mutex<Option<Zeroizing<Vec<u8>>>>>,` is_some()

## Motivation

Users should have an option to know if they still need to set the password, so they don't get an error because it's missing when they do an action

## Requirements

1. Add new public method to get if the key is set

No.

jyhi:
This PR attempts to fix #992 and #988, which include 3 improvements:

1. A method of `StrongholdAdapter` hinting if the key is still cached in the memory or not.
2. Aligning with what's been done in wallet-rs, removing `MessageType` from `message_interface`, and changing `ResponseType` to `Response`.
3. Purging secrets from the message interface wherever possible.


jyhi:
CIs failing oops

src/message_interface/message_handler.rs
```diff
@@ -66,22 +64,30 @@ impl ClientMessageHandler {
...
-        let response = match response {
-            Ok(r) => r,
-            Err(e) => ResponseType::Error(e),
+        // Zeroize secrets as soon as their missions are finished.
+        match &mut message.message_type {
+            MessageType::CallClientMethod(ClientMethod::StoreMnemonic { mnemonic, .. })
+            | MessageType::CallClientMethod(ClientMethod::MnemonicToHexSeed { mnemonic }) => {
+                mnemonic.zeroize();
+            }
+            _ => (),
```
Thoralf-M, line 83: Maybe add a comment, that it's not necessary for the methods that accept `SecretManagerDto` because you added `ZeroizeOnDrop` there

Thoralf-M (dismissed):

Cargo.toml
```diff
@@ -36,7 +36,7 @@ serde_json = { version = "1.0.68", default-features = false }
 thiserror = { version = "1.0.29", default-features = false }
 tokio = { version = "1.12.0", default-features = false, features = [ "macros", "rt-multi-thread", "time", "sync" ] }
 url = { version = "2.2.2", default-features = false, features = [ "serde" ] }
-zeroize = { version = "1.5.3", default-features = false }
+zeroize = { version = "1.5.3", default-features = false, features = ["zeroize_derive"] }
```
thibault-martinez:
```suggestion
zeroize = { version = "1.5.3", default-features = false, features = [ "zeroize_derive" ] }
```

Thoralf-M (approved):

thibault-martinez (approved):

PR #997, opened 2022-05-11
Add remainder to PreparedTransactionData

Thoralf-M:
# Description of change

Add remainder to PreparedTransactionData so we know which output is the remainder output by looking only on that
Required for offline signing support

## Type of change

## Change checklist

### DIFF
--- a/src/api/message_builder/input_selection/remainder.rs
+++ b/src/api/message_builder/input_selection/remainder.rs
@@ -32,7 +32,6 @@ pub(crate) async fn get_remainder_output<'a>(
     allow_burning: bool,
 ) -> Result<Option<RemainderData>> {
     log::debug!("[get_remainder]");
-    let mut remainder_data = None;
     let input_outputs = inputs.clone().map(|i| &i.output);
     let input_data = get_accumulated_output_amounts(std::iter::empty(), input_outputs.clone()).await?;
     let output_data = get_accumulated_output_amounts(std::iter::empty(), outputs.clone()).await?;
@@ -56,7 +55,7 @@ pub(crate) async fn get_remainder_output<'a>(
 
     let native_token_remainder = get_remainder_native_tokens(&input_native_tokens, &output_native_tokens)?;
     // Output possible remaining tokens back to the original address
-    if remainder_amount > 0 {
+    let remainder_data = if remainder_amount > 0 {
         let (remainder_addr, address_chain) = match remainder_address {
             // For provided remainder addresses we can't get the Chain
             Some(a) => (a, None),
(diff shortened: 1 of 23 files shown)


src/stronghold/secret.rs
```diff
@@ -129,6 +133,16 @@ impl SecretManage for StrongholdAdapter {
     }
 }
 
+#[async_trait]
+impl SecretManageExt for StrongholdAdapter {
+    async fn sign_transaction_essence(
+        &self,
+        prepared_transaction_data: &PreparedTransactionData,
+    ) -> crate::Result<Vec<UnlockBlock>> {
+        default_sign_transaction_essence(self, prepared_transaction_data).await
```
thibault-martinez: Use `self. default_sign_transaction_essence()` instead

src/secret/mnemonic.rs
```diff
@@ -86,6 +92,16 @@ impl SecretManage for MnemonicSecretManager {
     }
 }
 
+#[async_trait]
+impl SecretManageExt for MnemonicSecretManager {
+    async fn sign_transaction_essence(
+        &self,
+        prepared_transaction_data: &PreparedTransactionData,
+    ) -> crate::Result<Vec<UnlockBlock>> {
+        default_sign_transaction_essence(self, prepared_transaction_data).await
```
thibault-martinez: Same

src/api/message_builder/input_selection/remainder.rs
```diff
@@ -10,29 +10,34 @@ use bee_message::{
...
     outputs: impl Iterator<Item = &'a Output> + Clone,
     remainder_address: Option<Address>,
     byte_cost_config: &ByteCostConfig,
     allow_burning: bool,
-) -> Result<Option<Output>> {
+) -> Result<Option<RemainderData>> {
     log::debug!("[get_remainder]");
-    let mut remainder_output = None;
-    let input_data = get_accumulated_output_amounts(std::iter::empty(), inputs.clone()).await?;
+    let mut remainder_data = None;
```
thibault-martinez: Usually we'd prefer early returns or conditional returns to avoid mutation. You can actually make the last `if` to return either None or Some

src/stronghold/secret.rs
Thoralf-M: Doesn't work

thibault-martinez: Interesting, can you share the error ?

Thoralf-M:
```
no method named `default_sign_transaction_essence` found for reference `&mnemonic::MnemonicSecretManager` in the current scope
method not found in `&mnemonic::MnemonicSecretManager`
```
I removed it from 
```
impl<S> SecretManageExt for S
where
    S: SecretManage,
{
```
because it was conflicting with `sign_transaction_essence`, that's why it doesn't work

Changed it now

thibault-martinez: Ah I see. Just impl `default_sign_transaction_essence` in `SecretManage` instead of having it take a dynamic dispatched parameter.

thibault-martinez: Oh you did something similar while I was writing

thibault-martinez (approved):

samuel-rufi (approved):

PR #998, opened 2022-05-12
Remove node manager regexes

thibault-martinez:
## Change checklist

### DIFF
--- a/src/node_api/core/routes.rs
+++ b/src/node_api/core/routes.rs
@@ -29,7 +29,9 @@ impl Client {
     pub async fn get_info(&self) -> Result<NodeInfoWrapper> {
         let path = "api/v2/info";
 
-        self.node_manager.get_request(path, None, self.get_timeout()).await
+        self.node_manager
+            .get_request(path, None, self.get_timeout(), false, false)
+            .await
     }
 
     // Tangle routes.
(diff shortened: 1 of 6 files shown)


src/node_api/core/routes.rs
```diff
@@ -271,7 +275,9 @@ impl Client {
     pub async fn get_output(&self, output_id: &OutputId) -> Result<OutputResponse> {
         let path = &format!("api/v2/outputs/{}", output_id);
 
-        self.node_manager.get_request(path, None, self.get_timeout()).await
+        self.node_manager
+            .get_request(path, None, self.get_timeout(), true, true)
```
Thoralf-M: Not sure if needed here for quorum anymore, because with the inputs commitment a transaction would be invalid then if it's wrong

src/node_api/indexer/routes.rs
```diff
@@ -20,7 +20,8 @@ impl Client {
     pub async fn output_ids(&self, query_parameters: Vec<QueryParameter>) -> Result<Vec<OutputId>> {
         let route = "api/plugins/indexer/v1/outputs/basic";
 
-        self.get_output_ids_with_pagination(route, query_parameters).await
+        self.get_output_ids_with_pagination(route, query_parameters, false, false)
```
Thoralf-M: I would say for all indexer routes quorum should be true, since we only get unspent outputs and not spent outputs anymore they should return always the same thing

src/node_api/core/routes.rs
thibault-martinez: Right

Thoralf-M (approved):

pelumi-ajayi (approved):

PR #1002, opened 2022-05-16
Indexer API query parameters changes and checks

thibault-martinez:
## Change checklist

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -233,7 +233,7 @@ pub enum Error {
     #[error("Local time {0} doesn't match the time of the latest milestone timestamp: {1}")]
     TimeNotSynced(u32, u32),
     /// An indexer API request contains a query parameter not supported by the endpoint.
-    #[error("An indexer API request contains a query parameter not supported by the endpoint.")]
+    #[error("An indexer API request contains a query parameter not supported by the endpoint: {0}.")]
     UnsupportedQueryParameter(QueryParameter),
 }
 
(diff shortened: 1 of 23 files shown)


src/error.rs
```diff
@@ -230,6 +230,9 @@ pub enum Error {
     /// Local time doesn't match the time of the latest milestone timestamp
     #[error("Local time {0} doesn't match the time of the latest milestone timestamp: {1}")]
     TimeNotSynced(u32, u32),
+    /// An indexer API request contains a query parameter not supported by the endpoint.
+    #[error("An indexer API request contains a query parameter not supported by the endpoint.")]
+    UnsupportedQueryParameter,
```
Thoralf-M: Wouldn't it be better to also return which one is wrong?

thibault-martinez: Done

src/error.rs
```diff
@@ -232,7 +234,7 @@ pub enum Error {
     TimeNotSynced(u32, u32),
     /// An indexer API request contains a query parameter not supported by the endpoint.
     #[error("An indexer API request contains a query parameter not supported by the endpoint.")]
-    UnsupportedQueryParameter,
+    UnsupportedQueryParameter(QueryParameter),
```
Thoralf-M, line 237: I think for thiserror it should also be added in the message above with `{0}`

thibault-martinez: Ok but it means I have to impl Display for QueryParameter, are we fine with the impl just calling `to_query_string` ?

thibault-martinez: Done

Thoralf-M (approved):

samuel-rufi (approved):

PR #1005, opened 2022-05-17
Stronghold Improvements: Timeout and Unloading

### LINKED ISSUE #996: [Task]: Add a method to Stronghold to change the timeout
### Task description

Add a method to Stronghold to change the timeout for the timeout_task
If not running it should start it then

Also add a timeout to `StrongholdDto` which will set it upon creation of the `StrongholdSecretManager`

### Requirements

N/A

### Acceptance criteria

N/A

### Creation checklist

jyhi:
This PR adds timeout related methods to `StrongholdAdapter`. `get_timeout()` and `set_timeout()` are added to view and change the key clearing timeout. The task will be automatically spawned in `set_timeout()` if a new timeout is provided and key is still in the memory. `restart_key_timeout()` is also added as a shortcut of refreshing the count down by setting the same timeout again.

Fix #996.

### DIFF
--- a/src/secret/mod.rs
+++ b/src/secret/mod.rs
@@ -15,9 +15,9 @@ pub mod stronghold;
 /// Signing related types
 pub mod types;
 
+use std::{collections::HashMap, ops::Range, str::FromStr};
 #[cfg(feature = "stronghold")]
-use std::path::PathBuf;
-use std::{collections::HashMap, ops::Range, str::FromStr, time::Duration};
+use std::{path::PathBuf, time::Duration};
 
 use async_trait::async_trait;
 use bee_block::{
--- a/src/secret/mod.rs
+++ b/src/secret/mod.rs
@@ -17,7 +17,7 @@ pub mod types;
 
 #[cfg(feature = "stronghold")]
 use std::path::PathBuf;
-use std::{collections::HashMap, ops::Range, str::FromStr};
+use std::{collections::HashMap, ops::Range, str::FromStr, time::Duration};
 
 use async_trait::async_trait;
 use bee_block::{
@@ -165,6 +165,10 @@ impl TryFrom<&SecretManagerDto> for SecretManager {
                     builder = builder.password(password);
                 }
 
+                if let Some(timeout) = &stronghold_dto.timeout {
+                    builder = builder.timeout(Duration::from_secs(*timeout));
+                }
+
                 if let Some(snapshot_path) = &stronghold_dto.snapshot_path {
                     builder = builder.snapshot_path(PathBuf::from(snapshot_path));
                 }
@@ -191,6 +195,7 @@ impl From<&SecretManager> for SecretManagerDto {
             #[cfg(feature = "stronghold")]
             SecretManager::Stronghold(stronghold_dto) => Self::Stronghold(StrongholdDto {
                 password: None,
+                timeout: stronghold_dto.get_timeout().map(|duration| duration.as_secs()),
                 snapshot_path: stronghold_dto
                     .snapshot_path
                     .as_ref()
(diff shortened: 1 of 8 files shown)


Thoralf-M (dismissed):

src/stronghold/mod.rs
```diff
@@ -76,7 +78,7 @@ use crate::{Error, Result};
 #[builder(pattern = "owned", build_fn(skip))]
 pub struct StrongholdAdapter {
     /// A stronghold instance.
-    stronghold: Stronghold,
+    stronghold: Arc<Mutex<Stronghold>>,
```
Thoralf-M, line 73: Wouldn't a `RwLock` be better here?

Thoralf-M: Probably doesn't really help

jyhi: Maybe, but only `SecretManage` has 2 methods taking `&self`, while others take `&mut self`

Thoralf-M:
Shouldn't `clear_key()` also call `unload_stronghold_snapshot()`?

jyhi:
Oops, fixing.

src/stronghold/mod.rs
```diff
@@ -280,6 +280,11 @@ impl StrongholdAdapter {
         if let Some(mut key) = self.key.lock().await.take() {
             key.zeroize();
         }
+
+        // Unload Stronghold, but we can't do much about the errors.
+        if let Err(err) = self.unload_stronghold_snapshot().await {
```
Thoralf-M: This needs to be done before clearing the key, otherwise it will always fail

src/secret/mod.rs
```diff
@@ -191,6 +195,7 @@ impl From<&SecretManager> for SecretManagerDto {
             #[cfg(feature = "stronghold")]
             SecretManager::Stronghold(stronghold_dto) => Self::Stronghold(StrongholdDto {
```
thibault-martinez: Not part of your PR but the variable shouldn't be called `*_dto`, would be nice if you could change it.

src/stronghold/mod.rs
jyhi: Oops, yes.

Thoralf-M (dismissed):

Thoralf-M (dismissed):

src/stronghold/mod.rs
```diff
@@ -162,57 +168,61 @@ impl StrongholdAdapterBuilder {
         if let (Some(key), Some(Some(timeout))) = (&self.key, self.timeout) {
             let timeout_task = Arc::new(Mutex::new(None));
 
+            // Stronghold isn't optional, so we must initialize one here if it hasn't been supplied.
+            let stronghold = self.stronghold.unwrap_or_else(|| {
+                // XXX: we unwrap here.
```
thibault-martinez: Weird comment?

jyhi: It's for the next line; we're still on an older stable version of Stronghold, so we initialize `riker::ActorSystem` below, which returns a `Result`, but the `build()` method is not supposed to fail, so an `unwrap()` is left here.

It's actually a copy-pasta from the `Default` implementation (L118). To remove unwraps we can remove the `impl Default` and change `build() -> StrongholdAdapter` to `try_build() -> Result<StrongholdAdapter>`.

thibault-martinez: Yeah but I mean, this comment is basically useless. We can see that we are unwrapping, we would rather be expecting an explanation of why we are 100% sure this is not going to panic. And replace XXX with `PANIC`.

jyhi: Had a broader discussion and unwraps will be removed.

Thoralf-M (approved):

thibault-martinez (approved):

PR #1023, opened 2022-05-25
Feat/ledger nano

### LINKED ISSUE #873: [Task]: Update ledger signer to stardust
### Task description

Update ledger signer to stardust (ledger nano app needs to be updated first with trusted mode(signing only essence))

### Requirements

- [x] Generate addresses with different coin types
- [x] Sign transactions

### Acceptance criteria

Address generation with IOTA and SHIMMER coin type is working
transaction signing is working

### Creation checklist

- [x] I have assigned this task to the correct people

shufps:
# Description of change

Ledger nano signer (and blindsigning)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ rumqttc = { version = "0.12.0", default-features = false, features = [ "websocke
 once_cell = { version = "1.12.0", default-features = false, features = [ "std" ], optional = true }
 
 # ledger hardware wallets
-iota-ledger = { git = "https://github.com/iotaledger/ledger.rs", rev = "8611dbce8480ce82f6195705a57aa1311eb58547", default-features = false, optional = true }
+iota-ledger = { git = "https://github.com/iotaledger/ledger.rs", rev = "4a7a489f5c5899066c5ad27588690affef119b78", default-features = false, optional = true }
 
 # stronghold secret manager integration
 iota_stronghold = { version = "0.4.1", default-features = false, optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ rumqttc = { version = "0.12.0", default-features = false, features = [ "websocke
 once_cell = { version = "1.12.0", default-features = false, features = [ "std" ], optional = true }
 
 # ledger hardware wallets
-iota-ledger = { git = "https://github.com/iotaledger/ledger.rs", rev = "9e5029da7678f1cbf95862655b478409cbc881de", default-features = false, optional = true }
+iota-ledger = { git = "https://github.com/iotaledger/ledger.rs", rev = "8596552e78a6bac196c73518b0f26dc91bb39c8b", default-features = false, optional = true }
 
 # stronghold secret manager integration
 iota_stronghold = { version = "0.4.1", default-features = false, optional = true }
@@ -60,6 +60,7 @@ bee-ternary = { version = "0.6.0", default-features = false }
 
 [dev-dependencies]
 dotenv = { version = "0.15.0", default-features = false }
+env_logger = { version = "0.9.0", default-features = false }
 
 [features]
 default = [ "tls" ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ rumqttc = { version = "0.12.0", default-features = false, features = [ "websocke
 once_cell = { version = "1.12.0", default-features = false, features = [ "std" ], optional = true }
 
 # ledger hardware wallets
-iota-ledger = { git = "https://github.com/iotaledger/ledger.rs", rev = "4a7a489f5c5899066c5ad27588690affef119b78", default-features = false, optional = true }
+iota-ledger = { git = "https://github.com/iotaledger/ledger.rs", rev = "9f01355db6c5c3ea54726e1431a8354615894d5e", default-features = false, optional = true }
 
 # stronghold secret manager integration
 iota_stronghold = { version = "0.4.1", default-features = false, optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ rumqttc = { version = "0.12.0", default-features = false, features = [ "websocke
 once_cell = { version = "1.12.0", default-features = false, features = [ "std" ], optional = true }
 
 # ledger hardware wallets
-iota-ledger = { git = "https://github.com/iotaledger/ledger.rs", rev = "8596552e78a6bac196c73518b0f26dc91bb39c8b", default-features = false, optional = true }
+iota-ledger = { git = "https://github.com/iotaledger/ledger.rs", rev = "d5480d41f4fed11ff03a366149062bbcff0f44b8", default-features = false, optional = true }
 
 # stronghold secret manager integration
 iota_stronghold = { version = "0.4.1", default-features = false, optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,7 +60,6 @@ bee-ternary = { version = "0.6.0", default-features = false }
 
 [dev-dependencies]
 dotenv = { version = "0.15.0", default-features = false }
-env_logger = { version = "0.9.0", default-features = false }
 
 [features]
 default = [ "tls" ]
@@ -179,6 +178,10 @@ required-features = [ "mqtt" ]
 
 [[example]]
 name = "ledger_nano"
+required-features = [ "ledger_nano" ]
+
+[[example]]
+name = "ledger_nano_transaction"
 required-features = [ "ledger_nano" ]
 
 [[example]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ rumqttc = { version = "0.12.0", default-features = false, features = [ "websocke
 once_cell = { version = "1.12.0", default-features = false, features = [ "std" ], optional = true }
 
 # ledger hardware wallets
-iota-ledger = { git = "https://github.com/iotaledger/ledger.rs", rev = "9f01355db6c5c3ea54726e1431a8354615894d5e", default-features = false, optional = true }
+iota-ledger = { git = "https://github.com/iotaledger/ledger.rs", rev = "414c4b98294afc8863bed1b58c811a77105f7eea", default-features = false, optional = true }
 
 # stronghold secret manager integration
 iota_stronghold = { version = "0.4.1", default-features = false, optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ rumqttc = { version = "0.12.0", default-features = false, features = [ "websocke
 once_cell = { version = "1.12.0", default-features = false, features = [ "std" ], optional = true }
 
 # ledger hardware wallets
-iota-ledger = { git = "https://github.com/iotaledger/ledger.rs", rev = "d5480d41f4fed11ff03a366149062bbcff0f44b8", default-features = false, optional = true }
+iota-ledger = { git = "https://github.com/iotaledger/ledger.rs", rev = "8611dbce8480ce82f6195705a57aa1311eb58547", default-features = false, optional = true }
 
 # stronghold secret manager integration
 iota_stronghold = { version = "0.4.1", default-features = false, optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ rumqttc = { version = "0.12.0", default-features = false, features = [ "websocke
 once_cell = { version = "1.12.0", default-features = false, features = [ "std" ], optional = true }
 
 # ledger hardware wallets
-iota-ledger = { git = "https://github.com/iotaledger/ledger.rs", rev = "8c239b13b80e169aaa19cde9cd825e8444be0973", default-features = false, optional = true }
+iota-ledger = { git = "https://github.com/iotaledger/ledger.rs", rev = "9e5029da7678f1cbf95862655b478409cbc881de", default-features = false, optional = true }
 
 # stronghold secret manager integration
 iota_stronghold = { version = "0.4.1", default-features = false, optional = true }
--- a/examples/ledger_nano.rs
+++ b/examples/ledger_nano.rs
@@ -4,11 +4,14 @@
 //! cargo run --example ledger_nano --features=ledger_nano --release
 
 use iota_client::{
+    constants::{SHIMMER_COIN_TYPE, SHIMMER_TESTNET_BECH32_HRP},
     secret::{ledger_nano::LedgerSecretManager, SecretManager},
     Client, Result,
 };
 
 /// In this example we will create addresses with a ledger nano hardware wallet
+// To use the ledger nano simulator clone https://github.com/iotaledger/ledger-shimmer-app, run `git submodule init && git submodule update --recursive`,
+// then `./build.sh -m nanos|nanox|nanosplus -s` and use `true` in `LedgerSecretManager::new(true)`.
 
 #[tokio::main]
 async fn main() -> Result<()> {
@@ -19,11 +22,17 @@ async fn main() -> Result<()> {
         .finish()
         .await?;
 
-    let secret_manager = SecretManager::LedgerNano(LedgerSecretManager::new(false));
+    let ledger_nano = LedgerSecretManager::new(false);
+
+    println!("{:?}", ledger_nano.get_ledger_status().await);
+
+    let secret_manager = SecretManager::LedgerNano(ledger_nano);
 
     // Generate addresses with custom account index and range
     let addresses = client
         .get_addresses(&secret_manager)
+        .with_bech32_hrp(SHIMMER_TESTNET_BECH32_HRP)
+        .with_coin_type(SHIMMER_COIN_TYPE)
         .with_account_index(0)
         .with_range(0..2)
         .finish()
--- a/examples/ledger_nano.rs
+++ b/examples/ledger_nano.rs
@@ -14,7 +14,7 @@ use iota_client::{
 async fn main() -> Result<()> {
     // Create a client instance
     let client = Client::builder()
-        .with_node("https://api.alphanet.iotaledger.net:443")? // Insert your node URL here
+        .with_node("http://localhost:14265")? // Insert your node URL here
         .with_node_sync_disabled()
         .finish()
         .await?;
--- a/examples/ledger_nano_transaction.rs
+++ b/examples/ledger_nano_transaction.rs
@@ -19,7 +19,7 @@ async fn main() -> Result<()> {
         .finish()
         .await?;
 
-    let secret_manager = SecretManager::LedgerNano(LedgerSecretManager::new(false));
+    let secret_manager = SecretManager::LedgerNano(LedgerSecretManager::new(true));
 
     // Generate addresses with custom account index and range
     let addresses = client
--- a/examples/ledger_nano_transaction.rs
+++ b/examples/ledger_nano_transaction.rs
@@ -43,11 +43,7 @@ async fn main() -> Result<()> {
         .finish()
         .await?;
 
-    println!(
-        "Transaction sent with block-id: {}",
-        block.id()
-    );
-
+    println!("Transaction sent with block-id: {}", block.id());
 
     Ok(())
 }
--- a/examples/ledger_nano_transaction.rs
+++ b/examples/ledger_nano_transaction.rs
@@ -12,6 +12,8 @@ use iota_client::{
 
 #[tokio::main]
 async fn main() -> Result<()> {
+    env_logger::init();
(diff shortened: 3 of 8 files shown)


src/secret/ledger_nano.rs
```diff
@@ -279,6 +313,7 @@ impl LedgerSecretManager {
 
         log::info!("get_ledger");
         let (connected_, locked) = match iota_ledger::get_ledger(
+            0x107a, /* FIXME */
```
Thoralf-M: Does it make any difference for the status which coin type we use here?

shufps: I'm not sure, I have to check this :thinking:

examples/ledger_nano.rs
```diff
@@ -14,7 +14,7 @@ use iota_client::{
 async fn main() -> Result<()> {
     // Create a client instance
     let client = Client::builder()
-        .with_node("http://localhost:14265")? // Insert your node URL here
+        .with_node("https://api.alphanet.iotaledger.net:443")? // Insert your node URL here
```
Thoralf-M: Please undo

examples/ledger_nano_transaction.rs
```diff
@@ -0,0 +1,53 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! cargo run --example ledger_nano --features=ledger_nano --release
+
+use iota_client::{
+    secret::{ledger_nano::LedgerSecretManager, SecretManager},
+    Client, Result,
+};
+
+/// In this example we will create addresses with a ledger nano hardware wallet
```
Thoralf-M:
```suggestion
/// In this example we will create a transaction with a ledger nano hardware wallet
```

examples/ledger_nano_transaction.rs
```diff
@@ -0,0 +1,53 @@
+// Copyright 2021 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! cargo run --example ledger_nano --features=ledger_nano --release
```
Thoralf-M:
```suggestion
//! cargo run --example ledger_nano_transaction --features=ledger_nano --release
```

Thoralf-M: In the Cargo.toml we have 
```
[[example]]
name = "ledger_nano"
required-features = [ "ledger_nano" ]
```

```
[[example]]
name = "ledger_nano_transaction"
required-features = [ "ledger_nano" ]
```
should be added there

examples/ledger_nano_transaction.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    println!("List of generated public addresses:\n{:?}\n", addresses);
+
+    let block = client
+        .block()
+        .with_secret_manager(&secret_manager)
+        // Insert the output address and amount to spent. The amount cannot be zero.
+        .with_output(
+            // We generate an address from our seed so that we send the funds to ourselves
+            &client.get_addresses(&secret_manager).with_range(1..2).finish().await?[0],
```
Thoralf-M: You already generated addresses in line 25, no need to generate it again here

examples/ledger_nano_transaction.rs
```diff
@@ -0,0 +1,53 @@
...
+            1_000_000,
+        )?
+        .finish()
+        .await?;
+
+    println!(
+        "Transaction sent: https://explorer.alphanet.iotaledger.net/alphanet/block/{}",
```
Thoralf-M: This url should also stay "private"

src/secret/ledger_nano.rs
```diff
@@ -92,6 +94,32 @@ impl SecretManage for LedgerSecretManager {
     }
 }
 
+fn needs_blindsigning(prepared_transaction: &PreparedTransactionData) -> bool {
```
Thoralf-M: Please add a comment above what blindsigning is and why it's needed

src/secret/ledger_nano.rs
```diff
@@ -92,6 +94,32 @@ impl SecretManage for LedgerSecretManager {
...
+                        s.features().as_ref(),
+                    ) {
+                        // all fine, continue with next output
+                        continue;
+                    }
+                }
+                // not fine, return
+                return true;
+            }
+        } //_ => return true
```
Thoralf-M: Remove commented code

src/secret/ledger_nano.rs
```diff
@@ -101,154 +129,160 @@ impl SecretManageExt for LedgerSecretManager {
...
+
+        for input in &prepared_transaction.inputs_data {
+            let bip32_indices: Vec<u32> = match &input.chain {
+                Some(chain) => {
+                    chain
+                        .segments()
+                        .iter()
+                        // XXX: "ser32(i)". RTFSC: [crypto::keys::slip10::Segment::from_u32()]
+                        .map(|seg| u32::from_be_bytes(seg.bs()))
+                        .collect()
```
Thoralf-M, line 174: This shouldn't be needed, if we can update crypto.rs in ledger.rs to the same version

src/secret/ledger_nano.rs
```diff
@@ -101,154 +129,160 @@ impl SecretManageExt for LedgerSecretManager {
...
+        // sign
+        let signature_bytes = ledger.sign(input_len as u16)?;
+        let mut readable = &mut &*signature_bytes;
+        // unpack signature to unlockblocks
+        let mut unlock_blocks = Vec::new();
+        for _ in 0..input_len {
+            let unlock_block = Unlock::unpack_verified(&mut readable).map_err(|_| crate::Error::PackableError)?;
+            unlock_blocks.push(unlock_block);
+        }
+        Ok(unlock_blocks)
```
Thoralf-M: Remove `block`, only called `unlock` now since we renamed `Message` to `Block`

src/secret/ledger_nano.rs
```diff
@@ -101,154 +129,160 @@ impl SecretManageExt for LedgerSecretManager {
...
+
+        // sign
+        let signature_bytes = ledger.sign(input_len as u16)?;
+        let mut readable = &mut &*signature_bytes;
+        // unpack signature to unlockblocks
+        let mut unlock_blocks = Vec::new();
+        for _ in 0..input_len {
+            let unlock_block = Unlock::unpack_verified(&mut readable).map_err(|_| crate::Error::PackableError)?;
+            unlock_blocks.push(unlock_block);
+        }
```
Thoralf-M, line 322: I think I need to merge the signature unlocks with the reference unlocks here in the correct order, maybe add a `TODO` for me so I will do it in another PR then

examples/ledger_nano.rs
shufps: done

Thoralf-M: No, only in the other example :trollface:

Cargo.toml
```diff
@@ -63,6 +63,7 @@ bee-ternary = { version = "0.5.2", default-features = false }
 
 [dev-dependencies]
 dotenv = { version = "0.15.0", default-features = false }
+env_logger = { version = "0.9.0", default-features = false }
```
Thoralf-M: We already have `fern-logger` as dependency, you should be able to do the same things with it

src/secret/ledger_nano.rs
```diff
@@ -172,7 +176,7 @@ impl SecretManageExt for LedgerSecretManager {
 
         let ledger = iota_ledger::get_ledger(coin_type, bip32_account, self.is_simulator)?;
 
-        if needs_blindsigning(&prepared_transaction) {
+        if needs_blindsigning(prepared_transaction) {
```
shufps: wanted clippy :innocent:

src/secret/ledger_nano.rs
shufps: how would the alternative look like if we had the same version?

Thoralf-M: Ah I thought the ledger parts accepts `Chain`, but if that's not the case it needs to stay like this I guess

examples/ledger_nano.rs
shufps: done :see_no_evil:

src/secret/ledger_nano.rs
shufps: Hmm, I added the TODO but I'm not sure why the order should be off :thinking: The unlocks are returned in the order of the bip32 input indices and they have the same order as the inputs from the essence :thinking:

src/secret/ledger_nano.rs
shufps: ah yes, I think it makes a difference because when the app is split into iota and shimmer, getting the ledger object with coin_type 0x107a from the shimmer app will fail because the app won't support this :thinking:

shufps: I'll think about what we can change here

Thoralf-M: If the coin type is required we can add it as parameter to the method

src/secret/ledger_nano.rs
Thoralf-M: The order of these unlocks shouldn't be off, but if there are also other unlocks that are no signature, I might have to add them in between or just add the end, depending on the inputs order

src/secret/ledger_nano.rs
shufps: changed in: https://github.com/iotaledger/iota.rs/pull/1023/commits/3da725711c6d209e6a9474ce7df435b25623a217

src/secret/ledger_nano.rs
```diff
@@ -95,160 +101,209 @@ impl SecretManage for LedgerSecretManager {
...
+                        // all fine, continue with next output
+                        continue;
+                    }
+                }
+                // not fine, return
+                return true;
+            }
+        }
+    }
+    false
```
Thoralf-M: We still need to check the maximum amount of inputs/outputs based on the used ledger nano device

examples/ledger_nano_transaction.rs
```diff
@@ -0,0 +1,49 @@
+// Copyright 2021 IOTA Stiftung
```
Thoralf-M:
```suggestion
// Copyright 2022 IOTA Stiftung
```

src/secret/ledger_nano.rs
```diff
@@ -47,8 +53,8 @@ impl SecretManage for LedgerSecretManager {
     async fn generate_addresses(
         &self,
         // https://github.com/satoshilabs/slips/blob/master/slip-0044.md
-        // current ledger app only supports IOTA_COIN_TYPE
-        _coin_type: u32,
+        // current ledger app only supports IOTA_COIN_TYPE and SHIMMER_COIN_TYPE
```
Thoralf-M: And testnet coin type?

src/secret/ledger_nano.rs
```diff
@@ -95,160 +101,214 @@ impl SecretManage for LedgerSecretManager {
...
-                match &input.chain {
-                    Some(chain) => {
-                        let raw: Vec<u32> = chain
-                            .segments()
-                            .iter()
-                            // XXX: "ser32(i)". RTFSC: [crypto::keys::slip10::Segment::from_u32()]
-                            .map(|seg| u32::from_be_bytes(seg.bs()))
-                            .collect();
-                        (raw[0], raw[1])
+        let mut input_bip32_indices: Vec<iota_ledger::LedgerBIP32Index> = Vec::new();
```
Thoralf-M:
```suggestion
        let mut input_bip32_indices: Vec<LedgerBIP32Index> = Vec::new();
```

src/secret/ledger_nano.rs
```diff
@@ -95,160 +101,214 @@ impl SecretManage for LedgerSecretManager {
...
+                bip32_change: bip32_indices[3] | HARDENED,
+                bip32_index: bip32_indices[4] | HARDENED,
+            });
+        }
+
+        assert!(coin_type.is_some() && account_index.is_some());
+
+        // unwrap values
+        let coin_type = coin_type.unwrap() & !HARDENED;
+        let bip32_account = account_index.unwrap() | HARDENED;
```
Thoralf-M: Would be better to return errors here and not panic

samuel-rufi (dismissed):

src/secret/ledger_nano.rs
shufps: asserts are only for debug builds and only for things that obviously are very wrong and shouldn't happen at all. So imho it's not a case that should be handled externally and panic should be a clear hint that something in wallet.rs is broken :thinking:

shufps: if the asserts fail it would mean there are no inputs at all (totally wrong) or the chain of the inputs are different to each other (also totally wrong).

shufps: So I mean, it's not a case that actually ever could happen but if it does, something would be severly broken and therefore I wouldn't just return an error. But i wouldn't leave it completly unchecked ... assert is a bit of not ignoring^^

examples/ledger_nano_transaction.rs
shufps: done

src/secret/ledger_nano.rs
shufps: done

src/secret/ledger_nano.rs
shufps: done

src/secret/ledger_nano.rs
shufps: hmm but if you like, we could change it, of course :innocent:

Thoralf-M: Since when is it only in debug builds?
`Assertions are always checked in both debug and release builds, and cannot be disabled.` https://doc.rust-lang.org/std/macro.assert.html
And why can't it ever happen? I'm thinking about the offline example, where the inputs have the chain in the JSON, anyone could just provide something invalid there

shufps: oh, interesting - I think it was C++ or something like this that disables the asserts checking in release builds and I remembered wrong :see_no_evil: :thinking: and thank you for your explanations, then I'll change it :ok_hand:

src/secret/ledger_nano.rs
```diff
@@ -175,7 +177,9 @@ impl SecretManageExt for LedgerSecretManager {
             });
         }
 
-        assert!(coin_type.is_some() && account_index.is_some());
+        if coin_type.is_none() || account_index.is_none() {
+            return Err(crate::Error::NoInputs);
```
Thoralf-M, line 181:
```suggestion
            return Err(crate::Error::InvalidBIP32ChainData);
```

Thoralf-M (dismissed):

samuel-rufi (dismissed):

examples/ledger_nano_transaction.rs
thibault-martinez: Doesn't look done

src/secret/types.rs
```diff
@@ -61,15 +61,43 @@ pub struct LedgerApp {
...
+    LedgerNanoSPlus,
+}
+
+impl From<u8> for LedgerDeviceType {
+    fn from(device: u8) -> Self {
+        match device {
+            0 => Self::LedgerNanoS,
+            1 => Self::LedgerNanoX,
+            2 => Self::LedgerNanoSPlus,
+            _ => Self::LedgerNanoS,
```
thibault-martinez: Not a big fan of this, I'd rather impl TryFrom and error instead of returning a default. This may be totally fine, I'm missing the big picture, but better safe than sorry.

shufps: Hmm, the device type originally was intended for determining the size of the RAM buffer on the Ledger Nanos - in this case, it would have been ok because the Nano S has the smallest buffer of all devices. But since we now return the exact buffer size, we actually could remove it alltogether :man_shrugging: What do you think @Thoralf-M ?

Thoralf-M: I would still keep it, since it's additional info that could be helpful for someone

shufps: Ok, but should I change it to TryFrom? Would be okay for me :man_shrugging:

thibault-martinez: I think TryFrom would be less error prone

shufps: k, I'll do it - to finally get rid of the open PR :trollface:

src/secret/ledger_nano.rs
```diff
@@ -277,21 +353,119 @@ impl LedgerSecretManager {
...
+        let (connected_, locked, blindsigning_enabled, device) = match iota_ledger::get_app_config(&transport_type) {
+            Ok(config) => (
+                true,
+                // locked flag
+                config.flags & (1 << 0) != 0,
+                // blindsigning enabled flag
+                config.flags & (1 << 1) != 0,
+                match crate::secret::types::LedgerDeviceType::try_from(config.device) {
+                    Ok(d) => Some(d),
+                    Err(_) => None,
```
Thoralf-M: You can use `.ok()` instead

shufps: Thx, didn't know this!

Thoralf-M (dismissed):

src/secret/ledger_nano.rs
```diff
@@ -95,160 +114,217 @@ impl SecretManage for LedgerSecretManager {
...
+            };
+            // coin_type and account_index should be the same in each output
+            if (coin_type.is_some() && coin_type != Some(bip32_indices[0]))
+                || (account_index.is_some() && account_index != Some(bip32_indices[1]))
+            {
+                return Err(crate::Error::InvalidBIP32ChainData);
+            }
+
+            coin_type = Some(bip32_indices[1]);
+            account_index = Some(bip32_indices[2]);
```
thibault-martinez, line 186: Is this `1` and `2` ? Because in the `if` it was `0` and `1`

src/secret/ledger_nano.rs
```diff
@@ -95,160 +114,217 @@ impl SecretManage for LedgerSecretManager {
...
+        // if essence + bip32 input indices are larger than the buffer size or the essence contains
+        // features / types that are not supported blind signing will be needed
+        if blind_signing {
+            // prepare signing
+            log::debug!("[LEDGER] prepare_blind_signing");
+            log::debug!("[LEDGER] {:?} {:?}", input_bip32_indices, essence_hash);
+            ledger.prepare_blind_signing(input_bip32_indices, essence_hash)?;
+        } else {
+            // figure out the remainder address and bip32 index (if there is one)
+            let (has_remainder, remainder_address, remainder_bip32): (
```
thibault-martinez: `has_remainder` is useless since `remainder_address` is an Option

src/secret/ledger_nano.rs
```diff
@@ -95,160 +114,217 @@ impl SecretManage for LedgerSecretManager {
     }
 }
 
+/// needs_blind_signing
+/// the Ledger Nano S(+)/X app can present the user a detailed view of the transaction before it
+/// is signed but only with BasicOutputs, without extra-features and if the Essence is not too large.
+/// If criteria are not met, blind signing is needed.
+/// This method finds out if we have to switch to blind signing mode.
+pub fn needs_blind_signing(prepared_transaction: &PreparedTransactionData, buffer_size: usize) -> bool {
```
thibault-martinez, line 122: Does this need to be public ?

Thoralf-M: I think we want to use it in wallet.rs so Firefly can know what it needs to display (full tx or just essence hash) for verification

src/secret/ledger_nano.rs
shufps: ups, nice catch ... I guess we only tried with one single input :see_no_evil:

src/secret/ledger_nano.rs
shufps: k, removed :blush:

src/secret/ledger_nano.rs
```diff
@@ -232,19 +231,18 @@ impl SecretManageExt for LedgerSecretManager {
...
                         },
                     )
                 }
-                None => (false, None, iota_ledger::LedgerBIP32Index::default()),
+                None => (None, iota_ledger::LedgerBIP32Index::default()),
             };
 
             let mut remainder_index = 0u16;
-            if has_remainder {
+            if remainder_address.is_some() {
```
Thoralf-M: With `if let Some(remainder_address) = remainder_addresse{` we don't need to unwrap later `remainder_address.unwrap()`

thibault-martinez: Yes please do this change as CI is complaining about it

shufps: okidoki, I'll change it now :blush:

shufps: https://github.com/iotaledger/iota.rs/pull/1023/commits/5d4a31b7b2cf99c369799460ced792c1745f62bd

Thoralf-M (approved):

thibault-martinez (approved):

PR #1035, opened 2022-05-30
Improve `offline_signing` and `outputs` examples.

samuel-rufi:


examples/outputs/alias.rs
```diff
@@ -26,14 +27,16 @@ use iota_client::{
...
     let client = Client::builder()
         .with_node("http://localhost:14265")?
         .with_node_sync_disabled()
         .finish()
         .await?;
 
-    // This example uses dotenv, which is not safe for use in production
-    // Configure your own seed in ".env". Since the output amount cannot be zero, the seed must contain non-zero balance
+    // This example uses dotenv, which is not safe for use in production!
+    // Configure your own seed in the ".env" file. Since the output amount cannot be zero, the seed must contain
```
Thoralf-M:
```suggestion
    // Configure your own mnemonic in the ".env" file. Since the output amount cannot be zero, the seed must contain
```
I will try to remove the seed parts from most things, because it will be better to use mnemonics

examples/outputs/all_automatic_input_selection.rs
```diff
@@ -31,15 +32,17 @@ use primitive_types::U256;
...
         .with_node("http://localhost:14265")?
         .with_node_sync_disabled()
         .with_default_logger()?
         .finish()
         .await?;
 
-    // This example uses dotenv, which is not safe for use in production
-    // Configure your own seed in ".env". Since the output amount cannot be zero, the seed must contain non-zero balance
+    // This example uses dotenv, which is not safe for use in production!
+    // Configure your own seed in the ".env" file. Since the output amount cannot be zero, the seed must contain
```
Thoralf-M: Would also use mnemonic here

samuel-rufi: Good idea!

Thoralf-M (approved):

thibault-martinez (approved):

PR #1036, opened 2022-05-30
Rename outputs MQTT topics

samuel-rufi:
Fixes https://github.com/iotaledger/iota.rs/issues/1033

### DIFF
--- a/src/node_api/mqtt/types.rs
+++ b/src/node_api/mqtt/types.rs
@@ -171,9 +171,9 @@ impl Topic {
             r"^transactions/0x([a-f0-9]{64})/included-block$",
             // Output topics
             r"^outputs/([a-f0-9]{64})(\d{4})$",
-            r"^outputs/aliases/0x([a-f0-9]{40})$",
-            r"^outputs/nfts/0x([a-f0-9]{40})$",
-            r"^outputs/foundries/0x([a-f0-9]{52})$",
+            r"^outputs/alias/0x([a-f0-9]{40})$",
+            r"^outputs/nft/0x([a-f0-9]{40})$",
+            r"^outputs/foundry/0x([a-f0-9]{52})$",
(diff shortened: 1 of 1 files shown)


src/node_api/mqtt/types.rs
```diff
@@ -171,9 +171,9 @@ impl Topic {
             r"^transactions/0x([a-f0-9]{64})/included-block$",
             // Output topics
             r"^outputs/([a-f0-9]{64})(\d{4})$",
```
thibault-martinez, line 173: This one doesn't have a `basic` prefix like the core API?

samuel-rufi: Nope there is no prefix here <https://studio.asyncapi.com/?url=https://raw.githubusercontent.com/iotaledger/tips/stardust-event-api/tips/TIP-0028/event-api.yml>

samuel-rufi: But the Core API doesn't have a prefix too, only the Indexer API

thibault-martinez: I meant the indexer sorry

samuel-rufi: Hmm not sure now if this MQTT endpoint now also listens for outputs that are not "basic"

Thoralf-M: In https://github.com/iotaledger/inx-mqtt/commit/c06f2a2fffc0a825b06d8be38c1c874eb8893c7e it didn't get changed

Thoralf-M (approved):

thibault-martinez (approved):

src/node_api/mqtt/types.rs
samuel-rufi: okay let's keep it then

PR #1068, opened 2022-06-16
Use new stronghold version

thibault-martinez:


src/stronghold/secret.rs
```diff
@@ -36,31 +38,24 @@ impl SecretManage for StrongholdAdapter {
...
             vault_path: SECRET_VAULT_PATH.to_vec(),
             record_path: DERIVE_OUTPUT_RECORD_PATH.to_vec(),
         };
-        let hint = RecordHint::new(RECORD_HINT).unwrap();
 
         // Addresses to return.
         let mut addresses = Vec::new();
 
         for address_index in address_indexes {
             // Stronghold 0.4.1 is still using an older version of iota-crypto, so we construct a different one here.
```
Thoralf-M:
```suggestion
```

src/stronghold/secret.rs
```diff
@@ -95,15 +90,14 @@ impl SecretManage for StrongholdAdapter {
...
             vault_path: SECRET_VAULT_PATH.to_vec(),
             record_path: SEED_RECORD_PATH.to_vec(),
         });
         let derive_location = Location::Generic {
             vault_path: SECRET_VAULT_PATH.to_vec(),
             record_path: DERIVE_OUTPUT_RECORD_PATH.to_vec(),
         };
-        let hint = RecordHint::new(RECORD_HINT).unwrap();
 
         // Stronghold asks for an older version of [Chain], so we have to perform a conversion here.
```
Thoralf-M, line 98: Shouldn't be necessary anymore

Thoralf-M: Still needed

Cargo.toml
```diff
@@ -42,10 +42,7 @@ once_cell = { version = "1.12.0", default-features = false, features = [ "std" ]
 iota-ledger = { git = "https://github.com/iotaledger/ledger.rs", rev = "8c239b13b80e169aaa19cde9cd825e8444be0973", default-features = false, optional = true }
 
 # stronghold secret manager integration
-iota_stronghold = { version = "0.4.1", default-features = false, optional = true }
-riker = { version = "0.4.2", default-features = false, optional = true }
-# for stronghold
-crypto05 = { package = "iota-crypto", version = "0.5.1", default-features = false, features = [ "slip10" ], optional = true }
+iota_stronghold = { version = "0.6.0", default-features = false, features = [ "std" ], optional = true }
```
felsweg-iota: good to see, that our approach to reach `no_std` as default is integrated here, so `std` relevant features are explicitly opt-in

src/stronghold/mod.rs
```diff
@@ -112,16 +113,12 @@ impl StrongholdAdapterBuilder {
     /// Use an user-input password string to derive a key to use Stronghold.
     pub fn password(mut self, password: &str) -> Self {
         // Note that derive_builder always adds another layer of Option<T>.
-        self.key = Some(Arc::new(Mutex::new(Some(self::common::derive_key_from_password(
-            password,
-        )))));
+        self.key = Some(self::common::derive_key_from_password(password));
```
felsweg-iota: This is insecure. Even though a KDF will be used to store the secret inside an `Option`, the secret itself will remain in memory. I would recommend to re-use the `KeyProvider` data structure to store the key inside `StrongholdAdapter`. 
Further down below, the `KeyProvider` is already being used.

src/stronghold/mod.rs
```diff
@@ -135,24 +132,37 @@ impl StrongholdAdapterBuilder {
...
-                system,
-                client_path,
-                options,
-            )))
+            Stronghold::default()
         };
 
+        if let Some(key) = &self.key {
+            // TODO I don't think this.
+            let key_provider = KeyProvider::try_from((**key).clone())?;
```
felsweg-iota: See comment above. Better use the `KeyProvider` for storing keys.

src/stronghold/mod.rs
```diff
@@ -135,24 +132,37 @@ impl StrongholdAdapterBuilder {
...
+            let result = stronghold.load_client_from_snapshot(
+                PRIVATE_DATA_CLIENT_PATH,
+                &key_provider,
+                &SnapshotPath::from_path(&snapshot_path),
+            );
+            if let Err(iota_stronghold::ClientError::Inner(err_msg)) = result {
```
felsweg-iota: The `ClientError::Inner` variant is a compromise to support other kind of errors inside the `ClientError` enum, but we can discuss that, if there is a better solution.

src/stronghold/mod.rs
```diff
@@ -217,17 +227,16 @@ impl StrongholdAdapter {
...
-                    }
+        let result = self.read_stronghold_snapshot().await;
+        if let Err(err) = result {
+            // TODO: replace with actual error matching when updated to the new Stronghold version
+            // TODO: but the error is not an enum
+            if let crate::Error::StrongholdProcedureError(ref err_msg) = err {
+                if !err_msg.to_string().contains("IOError") {
+                    // If loading the snapshot failed but wasn't an IOError, then the password was incorrect and we
+                    // clear it
+                    *self.key.lock().await = None;
```
felsweg-iota: `EncryptionKey` is already wrapped in `Zeroize` which is good, maybe better to explicitly call `zeroize()` to make sure the data is cleared out immediately.

src/stronghold/mod.rs
```diff
@@ -436,28 +444,14 @@ impl StrongholdAdapter {
             return Err(Error::StrongholdKeyCleared);
         };
 
-        let snapshot_path = if let Some(path) = &self.snapshot_path {
-            path
-        } else {
-            return Err(Error::StrongholdSnapshotPathMissing);
-        };
+        // TODO: I don't like this
+        let key_provider = KeyProvider::try_from((**key).clone())?;
```
felsweg-iota: see above.

src/stronghold/mod.rs
```diff
@@ -478,38 +472,27 @@ impl StrongholdAdapter {
...
```
felsweg-iota: idem. see above.

src/stronghold/mod.rs
```diff
@@ -521,18 +504,6 @@ impl StrongholdAdapter {
...
-            .lock()
-            .await
-            .kill_stronghold(PRIVATE_DATA_CLIENT_PATH.to_vec(), false)
-            .await
-        {
-            ResultMessage::Ok(_) => Ok(()),
-            ResultMessage::Error(err) => Err(crate::Error::StrongholdProcedureError(err)),
-        }?;
-
         self.snapshot_loaded = false;
```
felsweg-iota, line 517: I wonder, if you really should go with a flag. One option would be assume Stronghold is being loaded and initialized. If access to certain secret paths fails, you can fall back to reload the Snapshot file from disk, or maybe I don't see the full picture here.

src/stronghold/mod.rs
```diff
@@ -542,7 +513,7 @@ impl StrongholdAdapter {
 /// The asynchronous key clearing task purging `key` after `timeout` spent in Tokio.
 async fn task_key_clear(
     task_self: Arc<Mutex<Option<JoinHandle<()>>>>,
-    stronghold: Arc<Mutex<Stronghold>>,
+    _stronghold: Arc<Mutex<Stronghold>>,
     key: Arc<Mutex<Option<Zeroizing<Vec<u8>>>>>,
```
felsweg-iota: If possible use `KeyProvider`

src/stronghold/secret.rs
```diff
@@ -12,11 +12,13 @@ use bee_block::{
...
-use iota_stronghold::{Location, ProcResult, Procedure, RecordHint, ResultMessage, SLIP10DeriveInput};
-use log::warn;
+use iota_stronghold::{
+    procedures::{self, Chain, KeyType, Slip10DeriveInput},
+    Location,
+};
 
 use super::{
-    common::{DERIVE_OUTPUT_RECORD_PATH, RECORD_HINT, SECRET_VAULT_PATH, SEED_RECORD_PATH},
+    common::{DERIVE_OUTPUT_RECORD_PATH, PRIVATE_DATA_CLIENT_PATH, SECRET_VAULT_PATH, SEED_RECORD_PATH},
```
felsweg-iota, line 21: `PRIVATE_DATA_CLIENT_PATH` is this some technical debt? Otherwise this could be config. I wonder, if this could also be stored as an non-contiguous data type at compile time, reassembled when needed :thinking:

felsweg-iota: On the other hand, this constant is obviously "public knowledge".

src/stronghold/secret.rs
```diff
@@ -36,35 +38,35 @@ impl SecretManage for StrongholdAdapter {
         _metadata: GenerateAddressMetadata,
     ) -> Result<Vec<Address>> {
         // Stronghold arguments.
-        let seed_location = SLIP10DeriveInput::Seed(Location::Generic {
+        let seed_location = Slip10DeriveInput::Seed(Location::Generic {
```
felsweg-iota: `Location` offers some convenience constructors like `Location::generic<V: Into<Vec<u8>>, R: Into<Vec<u8>>>(vault_path: V, record_path: R) -> Self` So you don't need to explicitly convert to `Vec<u8>`

src/stronghold/secret.rs
```diff
@@ -117,11 +118,11 @@ impl SecretManage for StrongholdAdapter {
                 .map(|seg| u32::from_be_bytes(seg.bs()))
                 .collect();
 
-            crypto05::keys::slip10::Chain::from_u32_hardened(raw)
+            Chain::from_u32_hardened(raw)
         };
 
         // Derive a SLIP-10 private key in the vault.
-        self.slip10_derive(chain, seed_location.clone(), derive_location.clone(), hint)
+        self.slip10_derive(chain, seed_location.clone(), derive_location.clone())
```
felsweg-iota: :party-parrot:

src/stronghold/secret.rs
```diff
@@ -142,139 +143,56 @@ impl SecretManage for StrongholdAdapter {
...
-        }
+            .load_client(PRIVATE_DATA_CLIENT_PATH)?
+            .execute_procedure(procedures::Slip10Derive { chain, input, output })?;
+
+        Ok(())
     }
 
     /// Execute [Procedure::Ed25519PublicKey] in Stronghold to get an Ed25519 public key from the SLIP-10 private key
     /// located in `private_key`.
     async fn ed25519_public_key(&self, private_key: Location) -> Result<[u8; 32]> {
```
felsweg-iota, line 162: I wonder, if we there is a way to easily export the Procedures as simple functions. But the effort here is limited, so may not be really an advantage.

felsweg-iota (commented):
Please check the inline comments for securely handling the secret for the Snapshot file.

src/stronghold/mod.rs
jyhi: You're correct, the flag was only a convenient design here.

src/stronghold/secret.rs
jyhi: They are technical debts. Normally the caller (library user) should set these (paths, filenames, ...) but historically it was hard-coded, so they were here again.

Thoralf-M (commented):
Building it with a wrong password returns `Error: StrongholdClient(ClientDataNotPresent)`, can't this be `WrongPassword` or something like that?

src/stronghold/db.rs
```diff
@@ -3,112 +3,92 @@
...
+        let key_provider = if let Some(key_provider) = &*locked_key_provider {
+            key_provider
         } else {
             return Err(Error::StrongholdKeyCleared);
         };
+        let buffer = key_provider.try_unlock()?;
+        let buffer_ref = buffer.borrow();
 
-        decrypt(&data, key).map(Some)
+        decrypt(&data, buffer_ref.deref()).map(Some)
```
felsweg-iota, line 41: :heart:

src/stronghold/db.rs
```diff
@@ -3,112 +3,92 @@
...
             .await
-            .write_to_store(location, encrypted_value, None)
-            .await;
-
-        if let ResultMessage::Error(err) = status {
-            return Err(Error::StrongholdProcedureError(err));
-        }
-
-        Ok(old_value)
+            .get_client(PRIVATE_DATA_CLIENT_PATH)?
```
felsweg-iota, line 62: I wonder, if this would be easier to actually have each object dereferenced explicitly.

felsweg-iota (dismissed):
Looks good to me. So good to see the `KeyProvider` extends its existence outside the scope of Stronghold :blush: . May it be of good use!

src/stronghold/common.rs
```diff
@@ -28,19 +23,24 @@ pub(super) const DERIVE_OUTPUT_RECORD_PATH: &[u8] = b"iota-wallet-derived";
 /// Filename to the Stronghold vault.
 ///
 /// The value has been hard-coded historically.
-pub(super) const STRONGHOLD_FILENAME: &str = "wallet.stronghold";
+pub(super) const _STRONGHOLD_FILENAME: &str = "wallet.stronghold";
```
Thoralf-M:
```suggestion
```

src/stronghold/mod.rs
```diff
@@ -209,25 +222,26 @@ impl StrongholdAdapter {
...
-                        // clear it
-                        *self.key.lock().await = None;
-                        return Err(err);
-                    }
+        let result = self.read_stronghold_snapshot().await;
+        if let Err(err) = result {
+            // TODO: replace with actual error matching when updated to the new Stronghold version
+            // TODO: but the error is not an enum
+            if let crate::Error::StrongholdProcedureError(ref err_msg) = err {
+                if !err_msg.to_string().contains("IOError") {
```
Thoralf-M: Shouldn't we get a `SnapshotFileMissing` error now?

Thoralf-M: https://github.com/iotaledger/stronghold.rs/commit/27c20730f9ab7ab6ab3807bdc1fac31e39d80dc0

thibault-martinez: Yes, I changed the one from `try_build` but forgot this one, will change

src/stronghold/mod.rs
```diff
@@ -542,23 +528,14 @@ impl StrongholdAdapter {
 /// The asynchronous key clearing task purging `key` after `timeout` spent in Tokio.
 async fn task_key_clear(
     task_self: Arc<Mutex<Option<JoinHandle<()>>>>,
-    stronghold: Arc<Mutex<Stronghold>>,
-    key: Arc<Mutex<Option<Zeroizing<Vec<u8>>>>>,
+    _stronghold: Arc<Mutex<Stronghold>>,
```
Thoralf-M: Can be removed now if we don't need to call a method to kill it anymore

src/stronghold/secret.rs
```diff
@@ -36,35 +38,29 @@ impl SecretManage for StrongholdAdapter {
...
-                address_index,
-            ]);
+            let chain = Chain::from_u32_hardened(vec![44u32, coin_type, account_index, internal as u32, address_index]);
+
+            let stronghold_client = self.stronghold.lock().await.get_client(PRIVATE_DATA_CLIENT_PATH)?;
 
             // Derive a SLIP-10 private key in the vault.
-            self.slip10_derive(chain, seed_location.clone(), derive_location.clone(), hint)
-                .await?;
+            stronghold_client.execute_procedure(procedures::Slip10Derive {
```
Thoralf-M: If we call the procedure like this now, we can remove `slip10_derive()` in secret.rs? Same for `ed25519_public_key()`

src/stronghold/mod.rs
```diff
@@ -270,9 +289,7 @@ impl StrongholdAdapter {
         }
```
Thoralf-M, line 281: keys_to_re_encrypt should be removed now and instead get the keys from stronghold itself, a method for that was added here https://github.com/iotaledger/stronghold.rs/pull/359

src/stronghold/secret.rs
thibault-martinez: I think I actually prefer the methods so maybe I'll replace the procedures with them

src/stronghold/mod.rs
thibault-martinez: Done

samuel-rufi (dismissed):

src/stronghold/mod.rs
```diff
@@ -209,25 +222,24 @@ impl StrongholdAdapter {
...
-                        // clear it
-                        *self.key.lock().await = None;
-                        return Err(err);
-                    }
+        let result = self.read_stronghold_snapshot().await;
+        if let Err(err) = result {
+            if let Error::StrongholdClient(ref err_msg) = err {
+                // Matching the error string is not ideal but stronhold doesn't wrap the error types at the moment.
+                if err_msg.to_string().contains("XCHACHA20-POLY1305") {
+                    // The the password was incorrect so we clear it.
```
Thoralf-M:
```suggestion
                    // The password was incorrect so we clear it.
```

src/stronghold/secret.rs
```diff
@@ -36,31 +38,17 @@ impl SecretManage for StrongholdAdapter {
         _metadata: GenerateAddressMetadata,
     ) -> Result<Vec<Address>> {
         // Stronghold arguments.
```
Thoralf-M, line 49: `Error: StrongholdProcedureError(Engine(FatalEngineError("vault `VaultId(TcAMhLekJE6GhvWh6qTYL-T2DKgaUKz7)` does not exist")))`
We get this error now if the key got cleared, and we then want to generate an address, maybe add the following?
```suggestion

        if !self.is_key_available().await {
            return Err(Error::StrongholdKeyCleared);
        }

        // Stronghold arguments.
```

thibault-martinez: Sure

Thoralf-M (dismissed):

src/stronghold/mod.rs
```diff
@@ -135,40 +134,54 @@ impl StrongholdAdapterBuilder {
...
 
+        if let Some(key_provider) = &self.key_provider {
+            let result = stronghold.load_client_from_snapshot(
+                PRIVATE_DATA_CLIENT_PATH,
+                key_provider,
+                &SnapshotPath::from_path(&snapshot_path),
+            );
+            if let Err(iota_stronghold::ClientError::SnapshotFileMissing(_)) = result {
+                stronghold.create_client(PRIVATE_DATA_CLIENT_PATH)?;
+                stronghold.write_client(PRIVATE_DATA_CLIENT_PATH)?;
```
felsweg-iota: `wirite_client` at this point would not be necessary to call, since `commit` will already write the just created client.

felsweg-iota (dismissed):
Looks good to me.

samuel-rufi (approved):

Thoralf-M (approved):

PR #1072, opened 2022-06-17
Java stardust support

samuel-rufi:


bindings/java/iota-client-java/gradlew
```diff
@@ -1,21 +1,5 @@
 #!/usr/bin/env sh
 
-#
-# Copyright 2015 the original author or authors.
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
```
Thoralf-M: Should this really be removed?

bindings/java/iota-client-java/native/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+name = "iota_client"
+crate-type = ["cdylib"]
+
+[dependencies]
+iota-client = { path = "../../../..", features = ["message_interface"] }
+jni = "0.19.0"
+serde_json = { version = "1.0.81", default-features = false }
+tokio = { version = "1.18.2", default-features = false, features = ["macros"] }
+once_cell = { version = "1.10.0", default-features = false }
+futures = { version = "0.3.21", default-features = false }
```
Thoralf-M: Order alphabetically

bindings/java/iota-client-java/native/src/main.rs
```diff
@@ -0,0 +1,4 @@
+// Copyright 2022 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+fn main() {}
```
Thoralf-M, line 4: Is this required?

bindings/java/iota-client-java/src/main/java/org/iota/apis/MiscellaneousApi.java
```diff
@@ -0,0 +1,131 @@
...
+        Float responsePayload = callBaseApi(new ClientCommand(ClientCommand.CommandType.CallClientMethod, "GetMinPoWScore")).getAsFloat();
+        return responsePayload;
+    }
+
+    public int getTipsInterval() throws ClientException {
+        Integer responsePayload = callBaseApi(new ClientCommand(ClientCommand.CommandType.CallClientMethod, "GetTipsInterval")).getAsInt();
+        return responsePayload;
+    }
+
+    public boolean isLocalPow() throws ClientException {
```
Thoralf-M: Why the different name? If it makes more sense to name things like this because they're local to the Client, then we should maybe also rename them in the Rust iota-client lib?

bindings/java/iota-client-java/src/main/java/org/iota/apis/NodeCoreApi.java
```diff
@@ -0,0 +1,224 @@
...
+        JsonObject responsePayload = (JsonObject) callBaseApi(new ClientCommand(ClientCommand.CommandType.CallClientMethod, "GetBlockMetadata", o));
+
+        return new BlockMetadata(responsePayload);
+    }
+
+    public Map.Entry<Output, OutputMetadata> getOutputWithMetadata(OutputId outputId) throws ClientException {
+        JsonObject o = new JsonObject();
+        o.addProperty("outputId", outputId.toString());
+
+        JsonObject responsePayload = (JsonObject) callBaseApi(new ClientCommand(ClientCommand.CommandType.CallClientMethod, "GetOutput", o));
```
Thoralf-M, line 90:
```suggestion
        JsonObject responsePayload = (JsonObject) callBaseApi(new ClientCommand(ClientCommand.CommandType.CallClientMethod, "GetOutputMetadata", o));
```

bindings/java/iota-client-java/src/test/java/org/iota/ApiTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.iota.types.ids.TransactionId;
+import org.iota.types.secret.GenerateAddressesOptions;
+import org.iota.types.secret.GenerateBlockOptions;
+import org.iota.types.secret.MnemonicSecretManager;
+import org.junit.jupiter.api.BeforeEach;
+
+public abstract class ApiTest {
+
+    protected static final String DEFAULT_DEVNET_NODE_URL = "https://api.alphanet.iotaledger.net";
+    protected static final String DEFAULT_DEVNET_FAUCET_URL = "https://faucet.alphanet.iotaledger.net";
```
Thoralf-M: Alphanet is a testnet and not a devnet :trollface:

bindings/java/iota-client-java/src/test/java/org/iota/ApiTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+    protected Block setUpTaggedDataBlock() throws ClientException {
+        return client.submitBlockPayload(new TransactionPayload("{ \"type\": 5, \"tag\": \"0x68656c6c6f20776f726c64\", \"data\": \"0x5370616d6d696e6720646174612e0a436f756e743a203037323935320a54696d657374616d703a20323032312d30322d31315431303a32333a34392b30313a30300a54697073656c656374696f6e3a203934c2b573\" }"));
+    }
+
+    protected Block setUpTransactionBlock() throws ClientException {
+        String address = client.generateAddresses(new MnemonicSecretManager(DEFAULT_DEVELOPMENT_MNEMONIC), new GenerateAddressesOptions().withRange(0, 1))[0];
+        requestFundsFromFaucet(address);
+        try {
+            Thread.sleep(1000 * 15);
```
Thoralf-M: Maybe move that into `requestFundsFromFaucet()`, since `requestFundsFromFaucet()` is called in multiple places

src/message_interface/client_method.rs
```diff
@@ -467,4 +472,40 @@ pub enum ClientMethod {
         /// Block
         block: BlockDto,
     },
+    /// Returns a transaction ID
```
Thoralf-M: Maybe also add that it's the hash of the provided transaction payload

src/message_interface/response.rs
```diff
@@ -138,6 +141,16 @@ pub enum Response {
     MnemonicHexSeed(String),
     /// The BLAKE2b-256 hash of the block bytes
     BlockId(BlockId),
+    /// The BLAKE2b-256 hash of the block bytes
+    TransactionId(TransactionId),
+    /// The BLAKE2b-256 hash of the block bytes
+    AliasId(AliasId),
+    /// The BLAKE2b-256 hash of the block bytes
+    NftId(NftId),
+    /// The BLAKE2b-256 hash of the block bytes
```
Thoralf-M: Wrong comments

bindings/java/iota-client-java/src/main/java/org/iota/types/Milestone.java
```diff
@@ -0,0 +1,15 @@
...
+
+    public Milestone(JsonObject jsonObject) {
+        super(jsonObject);
+    }
+
+    public Milestone(String jsonObject) {
+        super(jsonObject);
+    }
+
+}
```
Thoralf-M: For many types this is the same apart from the actual name, is there no less repetitive way with less code to do the same?

bindings/java/iota-client-java/gradlew
samuel-rufi: hm the one i'm using was automatically generated by gradle... but I will recheck

bindings/java/iota-client-java/src/test/java/org/iota/ApiTest.java
samuel-rufi: good point! or maybe we should even have this inside the Rust `requestFundsFromFaucet` ?

bindings/java/iota-client-java/src/main/java/org/iota/types/Milestone.java
samuel-rufi: Not that I'm familar with; there are no marcos like in Rust
Using abstract classes or interfaces to avoid repetitive way is the only way imo

bindings/java/iota-client-java/src/test/java/org/iota/ApiTest.java
samuel-rufi: Should we rename it to DEFAULT_TESTNET_NODE_URL and DEFAULT_TESTNET_FAUCET_URL ? 
we would need to update in the rust lib too

bindings/java/iota-client-java/native/Cargo.toml
samuel-rufi: done

bindings/java/iota-client-java/src/test/java/org/iota/ApiTest.java
Thoralf-M: > or maybe we should even have this inside the Rust requestFundsFromFaucet ?

Not sure about that, since one might also want to call it for multiple addresses and only then wait
For the tests you want to wait in each one, but there might be other use cases where it would be more annoying than helpful

bindings/java/iota-client-java/src/test/java/org/iota/ApiTest.java
Thoralf-M: Currently they should also not be added to the lib, there will be new urls when it's public for everyone

bindings/java/iota-client-java/src/main/java/org/iota/apis/MiscellaneousApi.java
samuel-rufi: In Java a good practice is to use the `is` prefix for boolean getters, so I thought it makes sense here to do (even if the Rust library doesn't) https://stackoverflow.com/a/5322666
In general for each binding it makes sense to follow its language specific best-practices for improved developer experience.. The "downside" is that the binding will differ a bit from the Rust library. What do you think?

Or are you talking about the "MiscellaneousApi" naming?

src/message_interface/response.rs
samuel-rufi: fixed

src/message_interface/client_method.rs
samuel-rufi: fixed

bindings/java/iota-client-java/src/test/java/org/iota/ApiTest.java
samuel-rufi: good point, changed it as you suggested above

bindings/java/iota-client-java/native/src/main.rs
samuel-rufi: fixed

bindings/java/iota-client-java/src/main/java/org/iota/apis/MiscellaneousApi.java
Thoralf-M: > it makes sense to follow its language specific best-practices for improved developer experience

That's true, I didn't know about that this is a good practice for Java
On the other hand having it aligned for all languages seems to be easier to maintain and could also help to make things you see in one language work for the other one, for example some examples or guides
Not sure what we should go for

bindings/java/iota-client-java/src/test/java/org/iota/ApiTest.java
samuel-rufi: fixed

bindings/java/iota-client-java/src/main/java/org/iota/apis/MiscellaneousApi.java
samuel-rufi: thats true.. i guess we should have a discussion about this question.
And maybe until then, do like the Rust library

samuel-rufi: changed it

bindings/java/iota-client-java/gradlew
samuel-rufi: fixed

bindings/java/iota-client-java/src/test/java/org/iota/ApiTest.java
Thoralf-M: But not pushed yet? I can't see the change

bindings/java/iota-client-java/src/main/java/org/iota/apis/MiscellaneousApi.java
```diff
@@ -0,0 +1,134 @@
...
+
+    public void storeMnemonic(SecretManager secretManager, String mnemonic) throws ClientException {
+        JsonObject o = new JsonObject();
+        o.add("secretManager", secretManager.getJson());
+        o.addProperty("mnemonic", mnemonic);
+
+        callBaseApi(new ClientCommand(ClientCommand.CommandType.CallClientMethod, "StoreMnemonic", o));
+    }
+
+    public Block postBlockPayload(BlockPayload payload) throws ClientException {
```
Thoralf-M: Not aligned with the message_interface, I think here renaming the part in the message interface would also be fine

bindings/java/iota-client-java/examples/SimpleProject/src/node_api_core/GetBlock.java
```diff
@@ -0,0 +1,30 @@
...
+    private static final String DEFAULT_TESTNET_NODE_URL = "http://localhost:14265";
+    private static ClientConfig config = new ClientConfig("{ \"nodes\": [\"" + DEFAULT_TESTNET_NODE_URL + "\" ], \"nodeSyncEnabled\": false}");
+
+    public static void main(String[] args) throws ClientException {
+        // Build the client.
+        Client client = new Client(config);
+
+        // Set up a block to later get it.
+        Block b = client.submitBlockPayload(new TaggedDataPayload("{\"type\":5,\"tag\":\"0x68656c6c6f20776f726c64\",\"data\":\"0x5370616d6d696e6720646174612e0a436f756e743a203037323935320a54696d657374616d703a20323032312d30322d31315431303a32333a34392b30313a30300a54697073656c656374696f6e3a203934c2b573\"}"));
+        BlockId blockId = client.postBlock(b);
```
Thoralf-M: The first one is already posting the block, so why post it again?

bindings/java/iota-client-java/examples/SimpleProject/src/node_api_core/GetBlockMetadata.java
```diff
@@ -0,0 +1,27 @@
...
+    private static final String DEFAULT_TESTNET_NODE_URL = "http://localhost:14265";
+    private static ClientConfig config = new ClientConfig("{ \"nodes\": [\"" + DEFAULT_TESTNET_NODE_URL + "\" ], \"nodeSyncEnabled\": false}");
+
+    public static void main(String[] args) throws ClientException {
+        // Build the client.
+        Client client = new Client(config);
+
+        // Set up a block to later get its metadata.
+        Block b = client.submitBlockPayload(new TaggedDataPayload("{\"type\":5,\"tag\":\"0x68656c6c6f20776f726c64\",\"data\":\"0x5370616d6d696e6720646174612e0a436f756e743a203037323935320a54696d657374616d703a20323032312d30322d31315431303a32333a34392b30313a30300a54697073656c656374696f6e3a203934c2b573\"}"));
+        BlockId blockId = client.postBlock(b);
```
Thoralf-M: Same here

bindings/java/iota-client-java/examples/SimpleProject/src/node_api_core/GetBlockRaw.java
```diff
@@ -0,0 +1,27 @@
...
+    private static final String DEFAULT_TESTNET_NODE_URL = "http://localhost:14265";
+    private static ClientConfig config = new ClientConfig("{ \"nodes\": [\"" + DEFAULT_TESTNET_NODE_URL + "\" ], \"nodeSyncEnabled\": false}");
+
+    public static void main(String[] args) throws ClientException {
+        // Build the client.
+        Client client = new Client(config);
+
+        // Set up a block to later get it.
+        Block b = client.submitBlockPayload(new TaggedDataPayload("{\"type\":5,\"tag\":\"0x68656c6c6f20776f726c64\",\"data\":\"0x5370616d6d696e6720646174612e0a436f756e743a203037323935320a54696d657374616d703a20323032312d30322d31315431303a32333a34392b30313a30300a54697073656c656374696f6e3a203934c2b573\"}"));
+        BlockId blockId = client.postBlock(b);
```
Thoralf-M: Same here

bindings/java/iota-client-java/examples/SimpleProject/src/node_api_core/PostBlock.java
```diff
@@ -0,0 +1,29 @@
...
+
+    private static final String DEFAULT_TESTNET_NODE_URL = "http://localhost:14265";
+    private static ClientConfig config = new ClientConfig("{ \"nodes\": [\"" + DEFAULT_TESTNET_NODE_URL + "\" ], \"nodeSyncEnabled\": false}");
+
+    public static void main(String[] args) throws ClientException {
+        // Build the client.
+        Client client = new Client(config);
+
+        // Set up a block.
+        Block b = client.submitBlockPayload(new TaggedDataPayload("{\"type\":5,\"tag\":\"0x68656c6c6f20776f726c64\",\"data\":\"0x5370616d6d696e6720646174612e0a436f756e743a203037323935320a54696d657374616d703a20323032312d30322d31315431303a32333a34392b30313a30300a54697073656c656374696f6e3a203934c2b573\"}"));
```
Thoralf-M: Here it's also already posting it, so the example is a little bit weird

bindings/java/iota-client-java/examples/SimpleProject/src/node_api_core/PostBlockRaw.java
```diff
@@ -0,0 +1,31 @@
...
+    private static final String DEFAULT_TESTNET_NODE_URL = "http://localhost:14265";
+    private static ClientConfig config = new ClientConfig("{ \"nodes\": [\"" + DEFAULT_TESTNET_NODE_URL + "\" ], \"nodeSyncEnabled\": false}");
+
+    public static void main(String[] args) throws ClientException {
+        // Build the client.
+        Client client = new Client(config);
+
+        // Set up a block.
+        Block block = client.submitBlockPayload(new TaggedDataPayload("{\"type\":5,\"tag\":\"0x68656c6c6f20776f726c64\",\"data\":\"0x5370616d6d696e6720646174612e0a436f756e743a203037323935320a54696d657374616d703a20323032312d30322d31315431303a32333a34392b30313a30300a54697073656c656374696f6e3a203934c2b573\"}"));
+        BlockId blockId = client.postBlock(block);
```
Thoralf-M: Same here

samuel-rufi: fixed

bindings/java/iota-client-java/examples/ExampleProject/src/node_api_core/GetBlockRaw.java
```diff
@@ -0,0 +1,24 @@
...
+public class GetBlockRaw {
+
+    private static final String DEFAULT_TESTNET_NODE_URL = "http://localhost:14265";
+    private static ClientConfig config = new ClientConfig("{ \"nodes\": [\"" + DEFAULT_TESTNET_NODE_URL + "\" ], \"nodeSyncEnabled\": false}");
+
+    public static void main(String[] args) throws ClientException {
+        // Build the client.
+        Client client = new Client(config);
+
+        // Set up a block for this example.
```
Thoralf-M:
```suggestion
        // Get a block id for which the block can be requested.
```

Thoralf-M (dismissed):

thibault-martinez (changes requested):
Let's not publish `bindings/java/iota-client-java/jar/iota-client-java-1.0-SNAPSHOT-linux-macos-windows-x86-64.jar` on the repository. We should either publish it as release artefact or recompiled by users but it's too big to be managed by Git. Same for the dll.

bindings/java/iota-client-java/examples/ExampleProject/src/node_api_core/GetBlockRaw.java
```diff
@@ -0,0 +1,24 @@
...
+
+    public static void main(String[] args) throws ClientException {
+        // Build the client.
+        Client client = new Client(config);
+
+        // Get a block id for which the block can be requested.
+        BlockId blockId = ExampleUtils.setUpBlockId(client);
+
+        // Get the block.
+        byte[] blockBytes = client.getBlockRaw(blockId);
```
thibault-martinez, line 22: Would be nice to still print the bytes in all raw examples

bindings/java/iota-client-java/src/main/java/org/iota/types/secret/GenerateAddressesOptions.java
```diff
@@ -0,0 +1,63 @@
...
+        o.addProperty("accountIndex", accountIndex);
+        o.add("range", range != null ? range.getAsJson() : null);
+        o.addProperty("internal", internal);
+        o.addProperty("bech32Hrp", bech32Hrp);
+        o.add("metadata", metadata != null ? metadata.getAsJson() : null);
+        return o;
+    }
+
+    static class GenerateAddressMetadata {
+        private boolean synching;
```
thibault-martinez: Use `syncing` instead, both are correct but in CS syncing is usually preferred

bindings/java/iota-client-java/examples/ExampleProject/src/node_api_core/GetBlockRaw.java
samuel-rufi: should we print them in hex?

thibault-martinez: Yeah it doesn't matter much, just that not printing anything is a bit confusing for the user, not knowing if it worked or not

bindings/java/iota-client-java/src/main/java/org/iota/types/secret/GenerateAddressesOptions.java
samuel-rufi: i never knew whats better to use, so thanks for the feedback! changed it

bindings/java/iota-client-java/src/main/java/org/iota/types/secret/GenerateAddressesOptions.java
```diff
@@ -0,0 +1,63 @@
...
+        o.addProperty("bech32Hrp", bech32Hrp);
+        o.add("metadata", metadata != null ? metadata.getAsJson() : null);
+        return o;
+    }
+
+    static class GenerateAddressMetadata {
+        private boolean syncing;
+
+        public GenerateAddressMetadata withSyncing(boolean synching) {
+            this.syncing = synching;
```
thibault-martinez: Rename, and do a search if there are no more

samuel-rufi: fixed

bindings/java/iota-client-java/examples/ExampleProject/src/node_api_core/GetBlockRaw.java
samuel-rufi: done

bindings/java/iota-client-java/src/main/java/org/iota/types/secret/GenerateAddressesOptions.java
```diff
@@ -0,0 +1,63 @@
...
+    public GenerateAddressesOptions withInternal(boolean internal) {
+        this.internal = internal;
+        return this;
+    }
+
+    public GenerateAddressesOptions withRange(int start, int end) {
+        range = new Range(start, end);
+        return this;
+    }
+
```
Thoralf-M, line 46: Don't we need a `withBech32Hrp` method here?

Thoralf-M (dismissed):

bindings/java/iota-client-java/examples/ExampleProject/src/WrongSeedConversionSecretManager.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private static final String DEFAULT_TESTNET_NODE_URL = "http://localhost:14265";
+    private static ClientConfig config = new ClientConfig("{ \"nodes\": [\"" + DEFAULT_TESTNET_NODE_URL + "\" ], \"nodeSyncEnabled\": false}");
+
+    public static void main(String[] args) throws ClientException {
+        testAddressDerivation();
+        migrate();
+    }
+
+    public static void testAddressDerivation() throws ClientException {
```
thibault-martinez: Is this an actual test ? I don't know Java but this looks like a regular class method ?

thibault-martinez (commented):
Do we have to push `bindings/java/iota-client-java/gradle/wrapper/gradle-wrapper.jar` ?

bindings/java/iota-client-java/src/main/java/org/iota/apis/BaseApi.java
```diff
@@ -0,0 +1,100 @@
...
+
+public class BaseApi {
+
+    protected ClientConfig clientConfig;
+
+    protected BaseApi(ClientConfig clientConfig) {
+        this.clientConfig = clientConfig;
+    }
+
+    static {
```
thibault-martinez, line 23: Should this really be done in the base API ?

bindings/java/iota-client-java/examples/ExampleProject/src/WrongSeedConversionSecretManager.java
samuel-rufi: good point.. it should actually be a test.. I think we should just move it to the actual library tests

bindings/java/iota-client-java/src/main/java/org/iota/apis/BaseApi.java
thibault-martinez: Wait actually what is the Base API? The name confused me, is that the message passing ?

bindings/java/iota-client-java/native/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    let client_command: String = env
+        .get_string(client_command)
+        .expect("Couldn't get java string!")
+        .into();
+
+    let message_handler = crate::block_on(async {
+        iota_client::message_interface::create_message_handler(Some(client_config.to_string())).await
+    })
+        .unwrap();
```
thibault-martinez, line 39: Is there no way to handle the unwraps in this file ?

thibault-martinez (dismissed):

Thoralf-M (dismissed):

thibault-martinez (dismissed):

thibault-martinez (approved):

Thoralf-M (approved):

PR #1087, opened 2022-06-27
Provide bytes for PostBlockRaw

samuel-rufi:


src/message_interface/client_method.rs
```diff
@@ -195,7 +195,7 @@ pub enum ClientMethod {
     /// Post block (raw)
     PostBlockRaw {
         /// Block
-        block: BlockDto,
+        block_bytes: Vec<u8>,
```
Thoralf-M, line 199:
```suggestion
        #[serde(rename = "blockBytes")]
        block_bytes: Vec<u8>,
```
the other bindings also need to be updated with this, either do in this PR or create an issue

Thoralf-M (approved):

thibault-martinez (approved):

PR #1091, opened 2022-06-28
Introduce SecretManagerDto::HexSeed variant

samuel-rufi:


Thoralf-M (approved):

pvdrz (approved):

src/secret/mod.rs
```diff
@@ -149,6 +149,8 @@ pub enum SecretManagerDto {
     LedgerNanoSimulator,
     /// Mnemonic
     Mnemonic(String),
+    /// Hex seed
+    HexSeed(String),
```
thibault-martinez, line 153: I think I would have called it just `Seed` tbh but whatever

thibault-martinez (approved):

src/secret/mod.rs
samuel-rufi: The method is called from_hex_seed too so for consistency i did the same

PR #1097, opened 2022-07-04
Add concurrency cancel-in-progress: true to workflows

Thoralf-M:
Add concurrency cancel-in-progress: true to workflows

### DIFF
--- a/.github/workflows/examine-python-bindings.yml
+++ b/.github/workflows/examine-python-bindings.yml
@@ -19,7 +19,7 @@ on:
     paths:
       - ".github/workflows/examine-python-bindings.yml"
       - ".github/actions/**"
-      - "bindings/python/native/**"
+      - "bindings/python/**"
       # Watch the Rust core too
       - "src/**"
       - "Cargo.toml"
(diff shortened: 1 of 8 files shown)


.github/workflows/examine-python-bindings.yml
```diff
@@ -19,7 +19,7 @@ on:
     paths:
       - ".github/workflows/examine-python-bindings.yml"
       - ".github/actions/**"
-      - "bindings/python/**"
+      - "bindings/python/native/**"
```
thibault-martinez: Not really sure we should do that. It would mean we wouldn't trigger actions if we change the Python code itself. Also, it's inconsistent with the push section.

Thoralf-M: Just added it because the workflow was running only for Python and I thought it's related to this, will undo

Thoralf-M:
```suggestion
      - "bindings/python/**"
```

thibault-martinez (approved):

samuel-rufi (approved):

PR #1117, opened 2022-07-11
Fetch node info from one of the nodes

pvdrz:
@Thoralf-M @thibault-martinez is this good enough?


thibault-martinez:
Closing as we won't do this for a while.

PR #1160, opened 2022-07-25
Fix testRetryUntilIncludedBlock + update example in README

samuel-rufi:


bindings/java/iota-client-java/src/test/java/org/iota/HighLevelApiTest.java
```diff
@@ -60,7 +57,15 @@ public void testRetryBlock() {
...
     @Test
     public void testRetryUntilIncludedBlock() throws ClientException {
-        LinkedHashMap<BlockId, Block> ret = client.retryUntilIncluded(client.getTips()[0], 1, 10);
+        SecretManager secretManager = new MnemonicSecretManager(client.generateMnemonic());
+        String[] addresses = client.generateAddresses(secretManager, new GenerateAddressesOptions().withRange(new Range(0, 2)));
+        requestFundsFromFaucet(addresses[0]);
+        GenerateBlockOptions.ClientBlockBuilderOutputAddress output = new GenerateBlockOptions.ClientBlockBuilderOutputAddress(addresses[1], Integer.toString(1000000));
+        Block b = client.generateBlock(secretManager, new GenerateBlockOptions().withOutput(output));
+        BlockId id = client.postBlock(b);
+        LinkedHashMap<BlockId, Block> ret = client.retryUntilIncluded(id, 1, 10);
```
Thoralf-M:
```suggestion
        LinkedHashMap<BlockId, Block> ret = client.retryUntilIncluded(id, 2, 15);
```
Maybe we should do this to be a little safer

samuel-rufi: good idea

Thoralf-M (approved):

PR #1195, opened 2022-08-14
Fix Wasm proof-of-work, time

cycraig:
# Description of change

Fixes several runtime and compilation errors for the `wasm32-unknown-unknown` target related to publishing and proof-of-work computations.

Notably, `finish_single_thread_pow` used outdated `Block` packing logic, causing publishing to fail completely on Wasm. It now goes through the same `NonceProvider` interface used to construct a `Block`, so it should be more maintainable.

This also consolidates some of the PoW logic but I made sure the non-Wasm logic is completely unchanged. It's just slightly cleaned up ~and now sets the workers in `promote_unchecked` and `reattach_unchecked` where it didn't before, not sure if that was intentional.~ Edit: ~reverted this change, but the Java bindings tests still fail with the following error, not sure why yet:~ https://github.com/iotaledger/iota.rs/runs/7830007231?check_suite_focus=true

Edit2: error was due to calling `ClientMiner::builder()`, which defaulted to `0` workers, instead of `ClientMinerBuilder::new()`. Un-reverted change and made `default` match `new`. The below errors no longer occur.

```
thread '<unnamed>' panicked at 'attempt to divide by zero', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/bee-pow-1.0.0-alpha.1/src/providers/miner.rs:164:28
thread '<unnamed>' panicked at 'Failed to join threads.: Any { .. }', /home/runner/work/iota.rs/iota.rs/src/api/block_builder/pow/mod.rs:66:28
```

The conditional PoW checks (single-threaded vs multi-threaded, whether or not to stop after the tips interval and fetch new parents) could be internalised further to a single `finish_pow` function, but I think being explicit might be preferred and it would mess with the public interface (e.g. `get_pow_provider` is public).

**NOTE: a new version of `bee-api-types` needs to be released with the changes in https://github.com/iotaledger/bee/pull/1495 (not sure if other `bee` crates need new versions too) for `iota.rs` to compile to the `wasm32-unknown-unknown` target.** I confirmed this PR works with the updated `bee-api-types` locally though.

### Changed 

- Change `Client::get_time_checked()` to use [instant](https://crates.io/crates/instant) on Wasm. It previously panicked with `"time not implemented on this platform"`.
- Change `Client::get_pow_provider()` to return a generic `NonceProvider` based on the target (`SingleThreadedMiner` on Wasm, otherwise `ClientMiner`). This fixes `retry_until_included`/`promote`/`reattach` calls for Wasm.
- Change `Client::get_pow_provider()` to set the worker count on `ClientMiner`.
- Change `ClientMinerBuilder::default()` implementation to match `new()` and default `local_pow` to `true`.
- Rename `finish_pow` to `finish_multi_threaded_pow`.
- Rename `finish_single_thread_pow` to `finish_single_threaded_pow`.

### Added

- Add `SingleThreadedMiner`.
- Add `SingleThreadedMinerBuilder`.

## Type of change

Technically a breaking change due to renaming public functions (`finish_pow`) and changing the return type of `Client::get_pow_provider()`.

Built on both targets:
- `cargo clippy`
- `cargo clippy --target wasm32-unknown-unknown`

Unit tests pass too.

I also ran examples using local Wasm bindings to confirm the runtime errors and panics no longer occur (coming in a future PR maybe, I followed the Node.js bindings approach and re-used its TypeScript definitions now, so hopefully it will be a plug-and-play replacement with a common base).

## Change checklist

- [x] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/src/api/block_builder/pow/mod.rs
+++ b/src/api/block_builder/pow/mod.rs
@@ -114,7 +114,7 @@ pub async fn finish_single_threaded_pow(client: &Client, payload: Option<Payload
 
         // The nonce defaults to 0 on errors (from the tips interval elapsing),
         // we need to re-run proof-of-work with new parents.
-        if block.nonce() == 0 && !local_pow {
+        if block.nonce() == 0 && local_pow {
             parent_blocks = client.get_tips().await?;
         } else {
             return Ok(block);
--- a/src/api/block_builder/pow/mod.rs
+++ b/src/api/block_builder/pow/mod.rs
@@ -3,23 +3,43 @@
 
 //! PoW functions
 
+#[cfg(target_family = "wasm")]
+pub mod wasm_miner;
+
 pub mod miner;
 
 use bee_block::{parent::Parents, payload::Payload, Block, BlockBuilder, BlockId};
+use bee_pow::providers::{NonceProvider, NonceProviderBuilder};
 use packable::PackableExt;
-#[cfg(not(target_family = "wasm"))]
-use {
-    crate::api::miner::ClientMinerBuilder,
-    bee_pow::providers::{miner::MinerCancel, NonceProviderBuilder},
-};
 #[cfg(target_family = "wasm")]
-use {bee_block::payload::OptionalPayload, packable::Packable};
-
-use crate::{api::miner::ClientMiner, Client, Error, Result};
-
-/// Does PoW with always new tips
+use wasm_miner::SingleThreadedMiner;
 #[cfg(not(target_family = "wasm"))]
-pub async fn finish_pow(client: &Client, payload: Option<Payload>) -> Result<Block> {
+use {crate::api::miner::ClientMiner, bee_pow::providers::miner::MinerCancel};
+
+use crate::{Client, Error, Result};
+
+/// Performs proof-of-work to construct a [`Block`].
+pub fn do_pow<P: NonceProvider>(
+    miner: P,
+    min_pow_score: f64,
+    payload: Option<Payload>,
+    parent_blocks: Vec<BlockId>,
+) -> Result<Block> {
+    let mut block = BlockBuilder::<P>::new(Parents::new(parent_blocks)?);
+    if let Some(p) = payload {
+        block = block.with_payload(p);
+    }
+    block
+        .with_nonce_provider(miner, min_pow_score)
+        .finish()
+        .map_err(Error::BlockError)
+}
+
+/// Performs multi-threaded proof-of-work.
+///
+/// Always fetches new tips after each tips interval elapses.
+#[cfg(not(target_family = "wasm"))]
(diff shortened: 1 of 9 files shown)


src/api/block_builder/pow/mod.rs
```diff
@@ -33,13 +53,12 @@ pub async fn finish_pow(client: &Client, payload: Option<Payload>) -> Result<Blo
         parent_blocks.dedup();
         let time_thread = std::thread::spawn(move || Ok(pow_timeout(tips_interval, cancel)));
         let pow_thread = std::thread::spawn(move || {
-            let mut client_miner = ClientMinerBuilder::new()
-                .with_local_pow(local_pow)
-                .with_cancel(cancel_2);
+            let mut client_miner = ClientMiner::builder().with_local_pow(local_pow).with_cancel(cancel_2);
```
Thoralf-M, line 56: Thanks for the PR!

I think this needs to stay (import also needs to be updated)
```suggestion
            let mut client_miner = ClientMinerBuilder::new().with_local_pow(local_pow).with_cancel(cancel_2);
```
the `attempt to divide by zero` doesn't happen in Java alone and should be fixed with this change

Can you also redo your change with the worker count for `promote_unchecked` and `reattach_unchecked`? Think it makes sense and shouldn't break anything

cycraig: Thanks! I didn't see that `ClientMinerBuilder::new` is different from the `Default` called by `::builder()`. 

Should the `Default` implementation be changed to always call `ClientMinerBuilder::new`?

```rust
    fn new() -> Self {
        Self {
            worker_count: num_cpus::get(),
            local_pow: false,
            cancel: MinerCancel::default(),
        }
    }
```

Thoralf-M: Yeah that would make sense, the default now doesn't make a lot of sense
I would actually also change the `local_pow` to `true` as default/in the `new()` method

Thoralf-M:
Ignore the failed Java tests, the nodes there are still not updated to the latest indexer changes

Thoralf-M (dismissed):

thibault-martinez (commented):
Would be nice if you could add your changes to the CHANGELOG please 🙏🏻

cycraig:
Added some `CHANGELOG.md` entries. Do I need to do something with the `.changes` directory?

Thoralf-M:
> Added some `CHANGELOG.md` entries. Do I need to do something with the `.changes` directory?

I don't think it's needed, the changes are not exposed to the nodejs binding and since it isn't using wasm the compilation also didn't fail there before

Thoralf-M (approved):

samuel-rufi (approved):

thibault-martinez (approved):

PR #1341, opened 2022-10-31
Fix nodejs package and path

Thoralf-M:
# Description of change

Fix nodejs package and path

## Type of change

Locally with `npm pack` and install

## Change checklist

### DIFF
--- a/client/bindings/nodejs/lib/bindings.ts
+++ b/client/bindings/nodejs/lib/bindings.ts
@@ -4,7 +4,7 @@
 import type { MessageHandler } from './MessageHandler';
 
 // @ts-ignore: path is set to match runtime transpiled js path
-import addon = require('../../index.node');
+import addon = require('../../build/Release/index.node');
 
 const { initLogger, sendMessage, messageHandlerNew, listen } = addon;
 
--- a/.changes/build-fix.md
+++ b/.changes/build-fix.md
@@ -1,0 +1,6 @@
+
+---
+"nodejs-binding": patch
+---
+
+Run tsc before publishing.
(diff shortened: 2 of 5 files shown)


client/bindings/nodejs/lib/bindings.ts
```diff
@@ -4,7 +4,7 @@
 import type { MessageHandler } from './MessageHandler';
 
 // @ts-ignore: path is set to match runtime transpiled js path
-import addon = require('../../index.node');
+import addon = require('../../build/Release/index.node');
```
thibault-martinez, line 7: Just checking, will this always be `Release`? Is it not dependent on how it's being built?

Thoralf-M: Yes, because of https://github.com/iotaledger/iota.rs/blob/c5b53f446fe523e7e8968d85d3161f47bbae33ce/client/bindings/nodejs/scripts/move-artifact.js#L4

thibault-martinez (approved):

client/bindings/nodejs/lib/bindings.ts
Thoralf-M: Or wait actually no, it will be build in another place, but then moved, so if you run it without the move script, then it's wrong

samuel-rufi (approved):

client/bindings/nodejs/lib/bindings.ts
Thoralf-M: But `npm i` will run it

thibault-martinez: So we good ?

Thoralf-M: I would say so, CI is fine and released package works also, would only be an issue if someone runs some other commands individually

PR #1346, opened 2022-10-31
Rework BlockBuilder nonce provider

### LINKED ISSUE #1322: Move client pow to iota_types

thibault-martinez:
## Change checklist


client/examples/node_api_core/04_post_block.rs
```diff
@@ -27,7 +28,11 @@ async fn main() -> Result<()> {
     // Get parents for the block.
     let parents = Parents::new(client.get_tips().await?)?;
     // Create the block.
-    let block = Block::build(parents).finish(min_pow_score)?;
+    let block = Block::build(parents).finish_nonce_provider(|bytes| {
```
thibault-martinez: Would be even better if we provide a function that does this, so they can just provide the function as parameter

Thoralf-M: You mean something like this?
```Rust
pub fn do_pow(bytes: &[u8], min_pow_score: u32) -> Result<u64>{
    let miner = MinerBuilder::new().with_num_workers(num_cpus::get()).finish();

    miner.nonce(bytes, min_pow_score)
}
```

thibault-martinez: Yes

Thoralf-M: And what about the other parameters like worker count or the cancel thing? Not needed?
Also I don't expect anyone really to use that, we can also just change the example to `client.block().finish().await?` and it will do the same (actually even better because it will update the tips if PoW takes too long)

client/CHANGELOG.md
```diff
@@ -24,15 +24,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `alias_id_to_bech32()` to utils;
+- `Client::finish_block_builder()`;
 
 ### Changed
 
 - Abort the sync nodes task through its handle instead of a oneshot channel;
 - Update network info also from unhealty nodes if ignore_node_health is set to true;
+- `finish_pow()` and `do_pow()` now accept optional parents and always do PoW, independent of the local PoW setting in Client;
```
Thoralf-M, line 36: We could make them private since `finish_block_builder()` does everything, but not sure if someone might still want to  have them

types/src/block/block.rs
```diff
@@ -57,22 +56,25 @@ impl<P: NonceProvider> BlockBuilder<P> {
...
-    pub fn with_nonce_provider(mut self, nonce_provider: P) -> Self {
-        self.nonce_provider = Some(nonce_provider);
+    pub fn with_nonce(mut self, nonce: u64) -> Self {
+        self.nonce = Some(nonce);
         self
     }
 
     /// Finishes the [`BlockBuilder`] into a [`Block`].
-    pub fn finish(self, min_pow_score: u32) -> Result<Block, Error> {
+    pub fn finish_nonce_provider<F: Fn(&[u8]) -> Result<u64, PoWError>>(
```
thibault-martinez: I actually don't really like having 2 finish, we should be able to error if no nonce or no nonce provider has been given, instead of putting a default 0 nonce which doesn't really make sense

thibault-martinez: Well, the only solution would have been to box the provider but we can't have a Clone requirement on boxed traits, so the builder couldn't be Clone anymore. Forget about it.

types/src/block/block.rs
```diff
@@ -307,7 +285,7 @@ pub mod dto {
     }
 
     impl Block {
-        fn _try_from_dto(value: &BlockDto) -> Result<BlockBuilder<u64>, DtoError> {
+        fn _try_from_dto(value: &BlockDto) -> Result<BlockBuilder, DtoError> {
```
samuel-rufi, line 288: is the _ prefix really needed here?

thibault-martinez: Yeah it's just a helper function to factorise the 2 other ones

thibault-martinez: Was already there anyway, bit outside of the context of this PR

samuel-rufi (dismissed):

samuel-rufi (approved):

Thoralf-M (approved):

types/src/block/block.rs
```diff
@@ -57,22 +57,21 @@ impl<P: NonceProvider> BlockBuilder<P> {
...
+    fn _finish(self) -> Result<(Block, Vec<u8>), Error> {
         verify_payload(self.payload.as_ref())?;
 
-        let mut block = Block {
+        let block = Block {
             protocol_version: self.protocol_version.unwrap_or(PROTOCOL_VERSION),
             parents: self.parents,
             payload: self.payload.into(),
-            nonce: 0,
+            nonce: self.nonce.unwrap_or(Self::DEFAULT_NONCE),
```
kwek20, line 74: Whats the point of a default nonce? Isnt setting an Option::None better? otherwise elsewhere we have to validate DEFAULT nonce vs block nonce again

types/src/block/block.rs
```diff
@@ -81,40 +80,19 @@ impl<P: NonceProvider> BlockBuilder<P> {
...
-
-        block.nonce = nonce_provider
-            .nonce(
-                &block_bytes[..block_bytes.len() - core::mem::size_of::<u64>()],
-                min_pow_score,
-            )
-            .unwrap_or(Self::DEFAULT_NONCE);
-
-        Ok(block)
+        Ok((block, block_bytes))
```
kwek20, line 83: As we can just call `pack_to_vec`, we return this here purely so we dont have to check the error?

pow/tests/miner.rs
```diff
@@ -2,27 +2,24 @@
...
 #[test]
-fn miner_provide() {
+fn miner_nonce() {
     let miner = MinerBuilder::new().with_num_workers(4).finish();
     let mut bytes = rand_bytes(256);
 
     let nonce = miner.nonce(&bytes[0..248], 4000).unwrap();
     bytes[248..].copy_from_slice(&nonce.to_le_bytes());
 
-    assert!(PoWScorer::new().score(&bytes) >= 4000f64);
```
kwek20, line 21: Aha the eternal battle PoW vs Pow

pow/src/miner.rs
```diff
@@ -77,31 +62,33 @@ pub struct MinerBuilder {
...
+    pub fn finish(self) -> Miner {
         Miner {
             num_workers: self.num_workers.unwrap_or(DEFAULT_NUM_WORKERS),
             cancel: self.cancel.unwrap_or_else(MinerCancel::new),
         }
     }
 }
 
-/// A nonce provider that mine nonces.
+/// A nonce miner
```
kwek20: Whats the reason for removing the Traits completely? This could still be into a Trait at least partually. I Dont like seeing 2 identical methods without a trait to prevent errors (nonce in this case)

pow/Cargo.toml
```diff
@@ -18,7 +18,11 @@ rustdoc-args = [ "--cfg", "doc_cfg" ]
 
 [dependencies]
 iota-crypto = { version = "0.15.3", default-features = false, features = [ "blake2b", "digest", "curl-p", "ternary_encoding" ] }
+num_cpus = { version = "1.14.0", default-features = false }
```
kwek20, line 21: I dont think this works on wasm! As were splitting them up, we should double check

thibault-martinez: Compiles fine on wasm for me

types/src/block/block.rs
thibault-martinez: What error ? We return it because we may use it later to compute the nonce and it's an expensive call

pow/src/miner.rs
thibault-martinez: The trait was causing more issues than solving and it wasn't really playing nice with the client. There is always one of these methods at a given time since they are for wasm or not wasm, so the trait is not really needed for one method.

types/src/block/block.rs
thibault-martinez: Yeah I'm not fully convinced about that either, I'll think about it

thibault-martinez: https://github.com/iotaledger/iota.rs/issues/1459

types/src/block/block.rs
kwek20:
```
let block_bytes = block.pack_to_vec();

if block_bytes.len() > Block::LENGTH_MAX {
    return Err(Error::InvalidBlockLength(block_bytes.len()));
}
Ok((block, block_bytes))
```

Its expensive indeed, so why make it at all, when we discard it half the time? `finish` discards and `finish_nonce` uses

thibault-martinez: But it is needed in both cases to check the error, it just happens we don't have any more use later for it if we don't compute the nonce

PR #1356, opened 2022-11-07
Input Selection

### LINKED ISSUE #1355: Implement new ISA

thibault-martinez:
## Change checklist


client/src/api/block_builder/input_selection/new/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+impl InputSelection {
+    pub fn filter(mut self, addresses: &[Address]) -> Self {
+        let addresses = addresses
+            .iter()
+            // TODO meh
+            .copied()
+            .chain(self.available_inputs.iter().filter_map(|input| match &input.output {
+                Output::Alias(output) => Some(Address::Alias(AliasAddress::from(
+                    output.alias_id().or_from_output_id(*input.output_id()),
+                ))),
```
Thoralf-M: I think here we need to over the outputs and check the state transition, if state index is the same as in the input, then it's a governor transition and then the address isn't unlocked
In all other cases the input selection would add them as state controller transition and then the address would be unlocked

client/src/api/block_builder/input_selection/new/burn.rs
```diff
@@ -0,0 +1,76 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use std::collections::{HashMap, HashSet};
+
+use primitive_types::U256;
+
+use crate::block::output::{AliasId, FoundryId, NftId, TokenId};
+
+/// A type to explicit what needs to be burned during input selection.
+#[derive(Default)]
```
Thoralf-M: Should this have Deserialize derived, so the bindings can provide it as JSON without us requiring an additional type? Serde rename to camelCase should then also be added

client/src/api/block_builder/input_selection/new/helper.rs
```diff
@@ -0,0 +1,38 @@
...
+        outputs
+            .iter()
+            .find_map(|output| {
+                if let Output::Alias(alias_output) = output {
+                    if *alias_output.alias_id() == alias_id {
+                        if alias_output.state_index() == alias_input.state_index() {
+                            // Governance transition.
+                            Some(Some(false))
+                        } else {
+                            // State transition.
```
Thoralf-M: To be valid the new state index needs to be +1, maybe we should also check this here?

client/src/api/block_builder/input_selection/new/mod.rs
```diff
@@ -0,0 +1,297 @@
...
+        let is_alias_state_transition = is_alias_state_transition(input, &outputs)?.unwrap_or(false);
+        let (required_address, _) =
+            input
+                .output
+                .required_and_unlocked_address(self.timestamp, input.output_id(), is_alias_state_transition)?;
+
+        match required_address {
+            Address::Alias(alias_address) => Ok(Some(Requirement::Alias(*alias_address.alias_id()))),
+            Address::Nft(nft_address) => Ok(Some(Requirement::Nft(*nft_address.nft_id()))),
+            _ => Ok(None),
```
Thoralf-M, line 76: Why are Ed25519 addresses not required?

client/src/api/block_builder/input_selection/new/mod.rs
```diff
@@ -0,0 +1,297 @@
...
+                _ => None,
+            }));
+
+        // self.available_inputs.retain(|input| input.can_be_unlocked_now(input, addresses, ???, self.time))
+
+        self
+    }
+
+    /// Selects inputs that meet the requirements of the outputs to satisfy the semantic validation of the overall
+    /// transaction. Also creates outputs if transitions are required.
```
Thoralf-M: Maybe something like this to include the remainder 
```suggestion
    /// transaction. Also creates a remainder output and chain transition outputs if required.
```

client/src/api/block_builder/input_selection/new/remainder.rs
```diff
@@ -0,0 +1,111 @@
...
+
+// Gets the remainder address from configuration of finds one from the inputs.
+// TODO should this also look for non-ed25519 addresses?
+// TODO should this also check available inputs ? not only selected.
+fn get_remainder_address(selected_inputs: &[InputSigningData], remainder_address: Option<Address>) -> Option<Address> {
+    if remainder_address.is_some() {
+        return remainder_address;
+    }
+
+    // TODO need to check timelock/expiration?
```
Thoralf-M:
```suggestion
    // TODO need to check timelock/expiration
```
:trollface:

client/src/api/block_builder/input_selection/new/remainder.rs
```diff
@@ -0,0 +1,111 @@
...
+    let mut output_native_tokens = get_native_tokens(outputs.iter().map(|output| &output.output))?;
+    let (minted_native_tokens, melted_native_tokens) = get_minted_and_melted_native_tokens(selected_inputs, outputs)?;
+
+    input_native_tokens.merge(minted_native_tokens)?;
+    output_native_tokens.merge(melted_native_tokens)?;
+    // TODO also merge burn
+
+    let native_tokens_diff = get_native_tokens_diff(&input_native_tokens, &output_native_tokens)?;
+
+    if inputs_sum > outputs_sum || native_tokens_diff.is_some() {
```
Thoralf-M: `if native_tokens_diff.is_some()`, but `inputs_sum <= outputs_sum` we would need to check if we can add more inputs, which is currently not happening, right?

client/src/api/block_builder/input_selection/new/requirement/base_token.rs
```diff
@@ -0,0 +1,160 @@
...
+        unlock_condition::UnlockCondition, AliasOutputBuilder, FoundryOutputBuilder, NftOutputBuilder, Output, Rent,
+    },
+    error::{Error, Result},
+    secret::types::InputSigningData,
+};
+
+// TODO checked operations ?
+
+pub(crate) fn base_token_sums(selected_inputs: &[InputSigningData], outputs: &[OutputInfo]) -> (u64, u64) {
+    let inputs_sum = selected_inputs.iter().map(|input| input.output.amount()).sum::<u64>();
```
Thoralf-M: The amount of SDRs is not subtracted here, if the required outputs with the SDR aren't created yet it might lead to some wrong result, because it could look as if more is "free" available even though it actually isn't

Looks fine how it's used now, but probably good to add a comment in case we want to use it also somewhere else

client/src/api/block_builder/input_selection/new/requirement/base_token.rs
```diff
@@ -0,0 +1,160 @@
...
+        // TODO this would be lowest amount of input strategy.
+        self.available_inputs
+            .sort_by(|left, right| left.output.amount().cmp(&right.output.amount()));
+
+        'overall: {
+            // 1. Basic with ED25519 address and nothing else
+            {
+                let inputs = self.available_inputs.iter().filter(|input| {
+                    if let Output::Basic(output) = &input.output {
+                        if let [UnlockCondition::Address(address)] = output.unlock_conditions().as_ref() {
```
Thoralf-M: I would say with an expired timelock and or expiration where our address is in the expiration unlock condition it should be used the same way

Is that what was meant with `// 2. Basic with ED25519 address and other things ????`?  Could be merged I think

client/src/api/block_builder/input_selection/new/requirement/base_token.rs
```diff
@@ -0,0 +1,160 @@
...
+                    .filter(|input| !input.output.is_basic())
+                    .peekable();
+
+                if inputs.peek().is_some() {
+                    for input in inputs {
+                        inputs_sum += input.output.amount();
+                        newly_selected_inputs.push(input.clone());
+                        newly_selected_ids.insert(*input.output_id());
+
+                        if inputs_sum >= outputs_sum {
```
Thoralf-M: This would be wrong if there is an input that requires us to create a storage deposit return output?

client/src/api/block_builder/input_selection/new/requirement/sender.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    /// Fulfills a sender requirement.
+    pub(crate) fn fulfill_sender_requirement(
+        &mut self,
+        address: Address,
+    ) -> Result<(Vec<InputSigningData>, Option<Requirement>)> {
+        match address {
+            Address::Ed25519(_) => self.fulfill_ed25519_address_requirement(address),
+            Address::Alias(alias_address) => match self.fulfill_alias_requirement(alias_address.into_alias_id()) {
+                Ok(res) => Ok(res),
```
Thoralf-M: This only checks if the alias is in the inputs, but not if it's a state transition

client/src/api/block_builder/input_selection/new/burn.rs
thibault-martinez: Sure

client/src/api/block_builder/input_selection/new/mod.rs
thibault-martinez: Well there are no further action like for alias and nft, what would you do ?

Thoralf-M: Yes, just think that from reading this function alone you don't know why they aren't required
Maybe renaming it to required_alias_nft_addresses or at least adding a comment would help

client/src/api/block_builder/input_selection/new/mod.rs
```diff
@@ -0,0 +1,297 @@
...
+            }));
+
+        // self.available_inputs.retain(|input| input.can_be_unlocked_now(input, addresses, ???, self.time))
+
+        self
+    }
+
+    /// Selects inputs that meet the requirements of the outputs to satisfy the semantic validation of the overall
+    /// transaction. Also creates a remainder output and chain transition outputs if required.
+    pub fn select(mut self) -> Result<(Vec<InputSigningData>, Vec<Output>)> {
```
Thoralf-M: What do you think about returning a `Selected` struct which has the inputs and outputs? Reading the tests I first had to look up what is what `selected.0` and `selected.1` makes this harder

client/tests/input_selection/burn.rs
```diff
@@ -0,0 +1,189 @@
...
+        .select();
+
+    assert!(matches!(
+        selected,
+        Err(Error::UnfulfillableRequirement(Requirement::Nft(nft_id))) if nft_id == nft_id_1
+    ));
+}
+
+#[test]
+fn burn_foundry_present() {
```
Thoralf-M, line 146: Reading this it would be interesting to have a test which mints some native tokens and at the same time burns some of them
Do you want to me to add this to the hackmd?

thibault-martinez: Yes please

client/src/api/block_builder/input_selection/new/mod.rs
thibault-martinez: Sure, will do

client/tests/input_selection/burn.rs
Thoralf-M: Done

client/src/api/block_builder/input_selection/new/requirement/mod.rs
```diff
@@ -29,7 +29,7 @@ pub enum Requirement {
     /// Foundry requirement.
     Foundry(FoundryId),
     /// Alias requirement.
-    Alias(AliasId),
+    Alias(AliasId, bool),
```
Thoralf-M, line 32: Pls add comment for the bool here

client/src/api/block_builder/input_selection/new/mod.rs
thibault-martinez: Done

client/tests/input_selection/alias_outputs.rs
```diff
@@ -0,0 +1,704 @@
...
+        alias_id_2,
+        BECH32_ADDRESS,
+        None,
+        None,
+        None,
+    )];
+
+    let selected = InputSelection::new(inputs.clone(), outputs.clone(), protocol_parameters)
+        .select()
+        .unwrap();
```
Thoralf-M: Shouldn't this fail because input and output has the same state index?

thibault-martinez: Aren't you allowed to do a governance transition for no specific reason ?

Thoralf-M: Oh right, already ignored this case because there is no reason for it :trollface:

client/src/api/block_builder/input_selection/new/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+
+    /// Selects inputs that meet the requirements of the outputs to satisfy the semantic validation of the overall
+    /// transaction. Also creates a remainder output and chain transition outputs if required.
+    pub fn select(mut self) -> Result<Selected> {
+        if self.available_inputs.is_empty() {
+            return Err(Error::NoInputsProvided);
+        }
+        if self.outputs.is_empty() {
+            return Err(Error::NoOutputsProvided);
+        }
```
Thoralf-M: Should this maybe be
```suggestion
        if self.outputs.is_empty() && self.burn.is_none() {
            return Err(Error::NoOutputsProvided);
        }
```
? I guess if someone wants to burn something, it would usually be the only thing one wants to do, so requiring to provide outputs seems like it would be annoying to use because one would need to provide an output with some random amount to an own address just because it can't be empty which could also lead to making it fail (not enough storage deposit for remainder or maybe larger amount than is freed from burning something else)

client/src/api/block_builder/input_selection/new/remainder.rs
```diff
@@ -0,0 +1,111 @@
...
+        remainder_builder = remainder_builder
+            .add_unlock_condition(UnlockCondition::Address(AddressUnlockCondition::new(remainder_address)));
+
+        if let Some(native_tokens) = native_tokens_diff {
+            remainder_builder = remainder_builder.with_native_tokens(native_tokens);
+        }
+
+        let remainder = remainder_builder.finish_output(protocol_parameters.token_supply())?;
+
+        // TODO should we always try to select enough inputs so the diff covers the deposit?
```
Thoralf-M: I think so, related to what I commented on L 79, because if we don't do it, it can often fail even if it would be possible

client/src/api/block_builder/input_selection/new/requirement/foundry.rs
```diff
@@ -0,0 +1,45 @@
...
+        &mut self,
+        foundry_id: FoundryId,
+    ) -> Result<(Vec<InputSigningData>, Option<Requirement>)> {
+        // Check if the requirement is already fulfilled.
+        if self
+            .selected_inputs
+            .iter()
+            .any(|input| is_foundry_with_id(&input.output, &foundry_id))
+        {
+            return Ok((Vec::new(), None));
```
Thoralf-M, line 32: Maybe this is done somewhere else, but I can't see it at the moment
Shouldn't this return the controlling alias as requirement?

thibault-martinez: This is solving the requirement, not detecting it, the controlling alias is found as requirement here https://github.com/iotaledger/iota.rs/blob/017ca875048d51d2138bf3daa6bfa96a742acfb6/client/src/api/block_builder/input_selection/new/requirement/mod.rs#L131

client/src/api/block_builder/input_selection/new/requirement/sender.rs
thibault-martinez: Fixed in https://github.com/iotaledger/iota.rs/pull/1356/commits/9a36f38eed61e50a6a944fc2714c443cd7306c05

client/src/api/block_builder/input_selection/new/mod.rs
thibault-martinez: Good point

thibault-martinez: Done in https://github.com/iotaledger/iota.rs/pull/1356/commits/de346947533d2c7f77efdd6fe90079d5a7ca0687

client/src/api/block_builder/input_selection/new/mod.rs
thibault-martinez: https://github.com/iotaledger/iota.rs/pull/1356/commits/89fb4b00d088317f9feec557685bf85adf08d6ad

client/src/api/block_builder/input_selection/new/remainder.rs
thibault-martinez: It is not possible at this point (in the remainder function), that `inputs_sum < outputs_sum`, and if it is equal then there is nothing to be done

client/src/api/block_builder/input_selection/new/requirement/foundry.rs
Thoralf-M: For what do we have the option to return a Requirement then?

thibault-martinez: Base token has to return a requirement, I just thought I would harmonise them in case they all need to return something at some point but it doesn't look like they do. I can remove.

Thoralf-M: Whatever you think is better, was just curious

thibault-martinez: Actually it's less error conversion if I keep it the way it is now

client/src/api/block_builder/input_selection/new/requirement/base_token.rs
thibault-martinez: We said we would filter for timelock and expiration at the beginning so we don't have to do it everywhere so yeah I guess you're right, it should be fine to do it all together

client/src/api/block_builder/input_selection/new/requirement/alias.rs
```diff
@@ -0,0 +1,142 @@
...
+            return Err(Error::UnfulfillableRequirement(Requirement::Alias(
+                alias_id,
+                state_transition,
+            )));
+        }
+
+        // If a state transition is not required, we can simply select the alias.
+        if !state_transition {
+            // Remove the output from the available inputs and return it, swap to make it O(1).
+            return Ok((vec![self.available_inputs.swap_remove(available_index.unwrap())], None));
```
samuel-rufi, line 112: is the unwrap() safe to do?

client/src/api/block_builder/input_selection/new/requirement/alias.rs
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        // At this point, a state transition is required so we need to verify that an alias output describing a
+        // governance transition was not provided.
+
+        let input = selected_input.unwrap_or_else(|| &self.available_inputs[available_index.unwrap()]);
```
samuel-rufi, line 119: is the unwrap() safe to do?

client/src/api/block_builder/input_selection/new/requirement/base_token.rs
```diff
@@ -0,0 +1,245 @@
...
+                        false
+                    }
+                });
+
+                for input in inputs {
+                    let sdruc = input
+                        .output
+                        .unlock_conditions()
+                        .and_then(|unlock_conditions| unlock_conditions.storage_deposit_return())
+                        .unwrap();
```
samuel-rufi: is the unwrap() safe to do?

client/src/api/block_builder/input_selection/new/requirement/native_tokens.rs
```diff
@@ -0,0 +1,178 @@
...
+                        .output
+                        .native_tokens()
+                        .map_or(false, |native_tokens| native_tokens.contains(diff.token_id()))
+                });
+
+                for input in inputs {
+                    amount += input
+                        .output
+                        .native_tokens()
+                        .unwrap()
```
samuel-rufi: is the unwrap() safe to do?

client/src/api/block_builder/input_selection/new/requirement/native_tokens.rs
```diff
@@ -0,0 +1,178 @@
...
+                        .get(diff.token_id())
+                        .unwrap()
```
samuel-rufi, line 151: is the unwrap() safe to do?

client/src/api/block_builder/input_selection/new/requirement/base_token.rs
thibault-martinez: I changed it. But instead of subtracting to inputs, I add to outputs. Achieves the same but I prefer it that way.

client/src/api/block_builder/input_selection/new/requirement/base_token.rs
thibault-martinez: It's been merged since

client/src/api/block_builder/input_selection/new/requirement/base_token.rs
thibault-martinez: Fixed

client/src/api/block_builder/input_selection/new/requirement/mod.rs
thibault-martinez: Done

client/src/api/block_builder/input_selection/new/remainder.rs
thibault-martinez: I took note of that, I'll do it in another PR

client/src/api/block_builder/input_selection/new/helper.rs
thibault-martinez: Took note and I'll do in another PR

client/src/api/block_builder/input_selection/new/requirement/sender.rs
```diff
@@ -0,0 +1,94 @@
...
+    block::address::Address,
+    error::{Error, Result},
+    secret::types::InputSigningData,
+};
+
+fn has_ed25519_address(input: &InputSigningData, address: &Address) -> bool {
+    // TODO could also be in state/governor?
+    if let Some(unlock_conditions) = input.output.unlock_conditions() {
+        if let Some(address_unlock_condition) = unlock_conditions.address() {
+            address_unlock_condition.address() == address
```
Thoralf-M, line 15: Shouldn't we check the expiration unlock condition before? Because if it's expired, then the address unlock condition will not be used

thibault-martinez: It'll be done once at the beginning, but it's not done yet, yes

Thoralf-M: But even if you check it at the beginning, you still want to include expired and not expired if your address is in the correct unlock condition?

client/src/api/block_builder/input_selection/new/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+pub(crate) struct OutputInfo {
+    pub(crate) output: Output,
+    pub(crate) provided: bool,
+}
+
+/// Working state for the input selection algorithm.
+pub struct InputSelection {
+    available_inputs: Vec<InputSigningData>,
+    // TODO option needed ?
+    required_inputs: Option<HashSet<OutputId>>,
```
samuel-rufi, line 38: if i remember, these come from the user that he absolutely want to feed the transaction right?

samuel-rufi (dismissed):

client/src/api/block_builder/input_selection/automatic.rs
```diff
@@ -177,7 +177,7 @@ impl<'a> ClientBlockBuilder<'a> {
                         Ok(r) => r,
                         // for these errors, just try again in the next round with more addresses which might have more
                         // outputs.
-                        Err(err @ crate::Error::NotEnoughBalance { .. }) => {
+                        Err(err @ crate::Error::InsufficientBaseTokenAmount { .. }) => {
```
Thoralf-M: Shouldn't it be BaseCoin? :trollface:

client/src/api/block_builder/input_selection/new/mod.rs
thibault-martinez: Yes

client/src/api/block_builder/input_selection/automatic.rs
thibault-martinez: Changed

client/src/api/block_builder/input_selection/new/remainder.rs
```diff
@@ -57,16 +59,45 @@ impl InputSelection {
         None
     }
 
+    pub(crate) fn remainder_amount(&self) -> Result<u64> {
```
Thoralf-M: Maybe rename, because this only returns a potential required remainder amount for an output with native tokens, but we could require/have a remainder amount without native tokens?

thibault-martinez: Is `required_remainder_amount` good for you ?

Thoralf-M: No, I think it's still missing the difference for the remainder amount without native tokens

thibault-martinez: Maybe `minimum_remainder_amount` makes sense instead.
Otherwise do you have a suggestion ?

Thoralf-M: But even then it only returns an amount if there are native tokens left
What I mean is that it completely misses the part when there is for example only 1 glow more on the inputs than on the outputs, nothing related to native tokens

thibault-martinez: Ah yeah ofc, sorry I'm in meetings so I didn't think this through, will do it later + add a test.

Thoralf-M: There is already `not_enough_storage_deposit_for_remainder()` so it seems to be handled fine already somewhere, just the naming of the functions sounds like it would also handle this case, but it doesn't

thibault-martinez: Yeah there is a test for that but I don't think there is a test when another available input should be selected to increase an already existing remainder amount that is less than the rent, and I don't think this case is currently handled at all

thibault-martinez: Can confirm the issue exist, will fix and then I think the name makes sense because it covers both

client/src/api/block_builder/input_selection/new/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+    pub fn timestamp(mut self, timestamp: u32) -> Self {
+        self.timestamp = timestamp;
+        self
+    }
+
+    // TODO should we somehow enforce using filter so we don't have to use can_be_unlocked_now later everywhere ?
+    /// Filters out the available inputs that
+    /// - can't be unlocked by the given addresses
+    /// - can't be unlocked now
+    pub fn filter(self, addresses: &[Address]) -> Self {
```
Thoralf-M, line 222: Can this be removed now?

thibault-martinez: No, but it's still unfinished

client/src/api/block_builder/input_selection/new/requirement/amount.rs
```diff
@@ -0,0 +1,276 @@
...
+                    self.available_inputs
+                        .retain(|input| !newly_selected_ids.contains(input.output_id()));
+
+                    // TODO explanation of Amount
+                    return Ok((newly_selected_inputs, Some(Requirement::Amount)));
+                }
+            }
+        }
+
+        'ici: {
```
Thoralf-M, line 225: What does this mean?

client/src/api/block_builder/input_selection/new/requirement/native_tokens.rs
```diff
@@ -0,0 +1,178 @@
...
+
+                                minted_native_tokens
+                                    .add_native_token(NativeToken::new(token_id, minted_native_token_amount)?)?;
+                            }
+                            Ordering::Less => {
+                                let melted_native_token_amount = input_foundry_simple_ts.circulating_supply()
+                                    - output_foundry_simple_ts.circulating_supply();
+
+                                melted_native_tokens
+                                    .add_native_token(NativeToken::new(token_id, melted_native_token_amount)?)?;
```
Thoralf-M, line 64: Should we maybe use the minted and melted fields instead? Should be fine the way it's now, if the outputs are correct, but one could be wrong and then we could return the error at this place here already

client/src/api/block_builder/input_selection/new/requirement/sender.rs
```diff
@@ -0,0 +1,94 @@
...
+
+use super::{InputSelection, Requirement};
+use crate::{
+    block::address::Address,
+    error::{Error, Result},
+    secret::types::InputSigningData,
+};
+
+fn has_ed25519_address(input: &InputSigningData, address: &Address) -> bool {
+    // TODO could also be in state/governor?
```
Thoralf-M, line 12: Yes, but we also need to know the transition type then

client/src/api/block_builder/input_selection/new/requirement/sender.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    /// Fulfills a sender requirement.
+    pub(crate) fn fulfill_sender_requirement(
+        &mut self,
+        address: Address,
+    ) -> Result<(Vec<InputSigningData>, Option<Requirement>)> {
+        match address {
+            Address::Ed25519(_) => self.fulfill_ed25519_address_requirement(address),
+            Address::Alias(alias_address) => {
+                match self.fulfill_alias_requirement(alias_address.into_alias_id(), true) {
```
Thoralf-M, line 77: Why is the state transition hardcoded to true here?

client/src/api/block_builder/input_selection/new/requirement/amount.rs
thibault-martinez: `hear` in French, don't bother, I'm removing it soon :trollface:

client/src/api/block_builder/input_selection/new/requirement/sender.rs
thibault-martinez: Because a state transition is needed to validate a sender

Thoralf-M: Sure, was just wondering what happens if it's a governance transition already, but I see it's checked in fulfill_alias_requirement() :+1:

client/src/api/block_builder/input_selection/new/requirement/amount.rs
thibault-martinez: Sorry I meant `here`

samuel-rufi (approved):

Thoralf-M (approved):
https://www.youtube.com/watch?v=KbDFCn29BGE

PR #1358, opened 2022-11-07
Verified OutputMetadata OutputId

thibault-martinez:
## Change checklist

### DIFF
--- a/client/src/api/block_builder/input_selection/mod.rs
+++ b/client/src/api/block_builder/input_selection/mod.rs
@@ -62,11 +62,6 @@ pub fn try_select_inputs(
         return Err(crate::Error::NoInputs);
     }
 
-    // Check that the data contains valid values to create output ids, because we can't properly return the error in all
-    // used places.
-    // TODO: validate data earlier, maybe on `InputSigningData` creation?
-    validate_input_output_ids(&mandatory_inputs, &additional_inputs)?;
-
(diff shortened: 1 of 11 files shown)


client/src/api/block_builder/input_selection/mod.rs
```diff
@@ -352,30 +349,51 @@ pub fn try_select_inputs(
...
         !mandatory_inputs.iter().any(|mandatory_input| {
-            input.output_metadata.transaction_id == mandatory_input.output_metadata.transaction_id
-                && input.output_metadata.output_index == mandatory_input.output_metadata.output_index
+            input.output_metadata.transaction_id() == mandatory_input.output_metadata.transaction_id()
+                && input.output_metadata.output_index() == mandatory_input.output_metadata.output_index()
         })
     });
 }
 
+// TODO is this still required?
```
thibault-martinez: ?

samuel-rufi (dismissed):

client/src/api/block_builder/input_selection/mod.rs
Thoralf-M: No can be removed now, it was there because the data from the node could have something which isn't a valid OutputId, but when we know that it's valid because it's required during the creation, then we can remove it

Thoralf-M (approved):

samuel-rufi (approved):

PR #1450, opened 2022-12-14
Update python version and name

Thoralf-M:
Update python version and name, otherwise it's not used in the released package

### DIFF
--- a/client/bindings/python/Cargo.toml
+++ b/client/bindings/python/Cargo.toml
@@ -1,8 +1,8 @@
 [workspace]
 
 [package]
-name = "iota-client-python"
-version = "0.2.0-alpha.3"
+name = "iota-client"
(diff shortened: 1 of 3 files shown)


samuel-rufi (dismissed):

samuel-rufi (dismissed):

client/bindings/python/Cargo.toml
```diff
@@ -1,8 +1,8 @@
 [workspace]
 
 [package]
-name = "iota-client-python"
```
thibault-martinez, line 4: I would like to keep the name like that for consistency but we can merge for now and find a solution later. There is already an issue to sort these things out.

thibault-martinez (approved):

samuel-rufi (approved):

PR #1454, opened 2022-12-16
Add coverage workflow

Thoralf-M:


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,76 @@
...
+        # Restore from outdated cache for speed
+        restore-keys: |
+          ubuntu-latest-nightly-cargo-index-${{ hashFiles('**/Cargo.lock') }}
+          ubuntu-latest-nightly-cargo-index-
+
+    - name: Install binutils and rustfilt
+      run: cargo install cargo-binutils rustfilt
+
+    - name: Run tests and collect coverage
+      working-directory: ./client
```
thibault-martinez: Can't we do coverage of the whole repository ? We surely want types to be covered as well

Thoralf-M: That's done, see https://coveralls.io/github/Thoralf-M/iota.rs
But when I moved the coverage.sh file to the root it returned 0 as coverage, so I moved it back there :shrug:

thibault-martinez: Oof that's going to trigger me

Thoralf-M: This was the run where it returned 0, do you see anything wrong there? https://github.com/Thoralf-M/iota.rs/actions/runs/3712563503/jobs/6294235150

thibault-martinez: Hmm no not really

thibault-martinez: I'll have a look at coverage.sh later, there has to be something with the fact that ./ is a workspace

client/coverage.sh
```diff
@@ -0,0 +1,44 @@
...
+rm -rf coverage
+mkdir coverage
+
+# Run tests with profiling instrumentation
+echo "Running instrumented unit tests..."
+RUSTFLAGS="-C instrument-coverage" LLVM_PROFILE_FILE="iota-client-%m.profraw" cargo +nightly test --all --all-features
+
+# Merge all .profraw files into "iota-client.profdata"
+echo "Merging coverage data..."
+cargo +nightly profdata -- merge ./iota-client-*.profraw -o iota-client.profdata
```
kwek20: probably here it tries to find this data but it cant when we have it in the workspace?

.github/workflows/coverage.yml
Thoralf-M: All is covered now https://coveralls.io/github/Thoralf-M/iota.rs

thibault-martinez (approved):

kwek20 (approved):

PR #1502, opened 2023-01-20
Build wheels for more python versions and aarch64-apple-darwin target

### LINKED ISSUE #1497: Build Python bindings for multiple Python versions
## Description

I tried installing the rc for this library from pip, but it wouldn't install as I use Python 3.11. So it would be better to build the package for multiple versions of find out another way to make it compatible.

No.

Thoralf-M:
# Description of change

Build wheels for more python versions and aarch64-apple-darwin target

Fixes https://github.com/iotaledger/iota.rs/issues/1497

## Type of change

In my fork https://github.com/Thoralf-M/iota.rs/actions/runs/3967181672/

### DIFF
--- a/.github/workflows/release-python-bindings.yml
+++ b/.github/workflows/release-python-bindings.yml
@@ -57,13 +57,12 @@ jobs:
         if: matrix.os == 'macos-latest' || matrix.os == 'windows-latest'
         uses: PyO3/maturin-action@v1
         with:
-          args: --release --out dist -m client/bindings/python/Cargo.toml
+          args: --release --out dist -m client/bindings/python/Cargo.toml -i python${{ matrix.python-version }}
 
       - name: Build wheels for --target aarch64-apple-darwin
         if: matrix.os == 'macos-latest'
         uses: PyO3/maturin-action@v1
         with:
-          container: off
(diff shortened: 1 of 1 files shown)


kwek20 (dismissed):

.github/workflows/release-python-bindings.yml
```diff
@@ -59,6 +59,13 @@ jobs:
         with:
           args: --release --out dist -m client/bindings/python/Cargo.toml
```
thibault-martinez: Only aarch64-apple-darwin is built for different versions ?

Thoralf-M: It should detect the installed version, so probably can be also removed from aarch64-apple-darwin, do you want me to change?
You can see that it detects the correct version in the logs here https://github.com/Thoralf-M/iota.rs/actions/runs/3967181672/jobs/6798844991
> Found CPython 3.8 at /opt/hostedtoolcache/Python/3.8.16/x64/bin/python
> Found CPython 3.9 at /opt/hostedtoolcache/Python/3.9.16/x64/bin/python

samuel-rufi (dismissed):

.github/workflows/release-python-bindings.yml
thibault-martinez: Better if it's consistent I think, yes

thibault-martinez (approved):

kwek20 (approved):

Issue #1522, opened 2023-01-26
[Bug]: Build Node.js Prebuilds fails


Thoralf-M:
### Issue description

https://github.com/iotaledger/iota.rs/actions/runs/4018292783/jobs/6903761126

Maybe we should also change the build workflow to only build for N-API the same way as in wallet.rs, because then only 3 binaries should be required

### Version

-

### Expected behaviour

Work

### Actual behaviour

Fails

### Can the issue reliably be reproduced?

Yes

### Duplicate declaration

- [X] I have searched the issues tracker this issue and there is none

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1556, opened 2023-02-06
feat: use N-API 6 builds

Tuditi:
# Description of change

Update the workflow and prebuild scripts to use N-API 6 builds for the nodeJS bindings.

https://github.com/iotaledger/iota.rs/issues/1522

## Type of change

Run the prebuild scripts locally

## Change checklist


.github/workflows/covector.yml
```diff
@@ -92,16 +92,6 @@ jobs:
         # tests on Windows 2019, until we can figure out a way to fix the problem.
         # NOTE: Using Ubuntu 18.04 to provide glibc compatibility. (#588)
         os: [ubuntu-18.04, ubuntu-latest, macos-latest, windows-2019]
```
Thoralf-M: Do we still need ubuntu-18.04 here?

rajivshah3: I believe we're using 20.04 on `wallet.rs` bindings now so it's probably safe to do that here too

.github/workflows/covector.yml
```diff
@@ -92,16 +92,6 @@ jobs:
...
-        node-version: ['14.x', '16.x', '18.x']
```
rajivshah3, line 95: I'm on mobile so I can't check the default version that `actions/setup-node` uses, but I think the step on line 106 needs to be updated if we remove this. On `wallet.rs` we use 16.x (although tbh, any LTS version of Node should work as long as the scripts don't break)

Tuditi: As far as I see, wallet.rs uses 14.x, but maybe it's worth updating that one to 16.x as well. I set it here to 16.x.

.github/workflows/covector.yml
Tuditi: It depends if we want to support older glibc versions. I updated it for now, but if users request it, we might need to roll back this change.

Thoralf-M (approved):

thibault-martinez (approved):

PR #1557, opened 2023-02-07
Use OutputMetadataDto instead of OutputMetadata

### LINKED ISSUE #1547: [Bug]: `InputSigningDataDto` incompatible with NodeJS binding
### Issue description

I'm exposing an API (`SecretManager::signature_unlock`) in the nodejs bindings and when testing that I ran into some issues. I wanted to validate that the behavior on the nodejs API is the same as in Rust. In Rust I create a `PreparedTransactionDataDto` and serialize that. The result is used in a nodejs test as the input for the above API to then assert that, given the same secret managers with the same mnemonic, they produce the same `Unlock` for a certain input of the transaction.

### Version

`a2d98bc0` (`develop` branch from yesterday).

### Expected behaviour

I expect to be able to serialize `PreparedTransactionDataDto` as JSON and use the result as an object in TS of type `IPreparedTransactionData`.

### Actual behaviour

Three errors/problems occur when doing that.

1. `PreparedTransactionDataDto` contains an `InputSigningDataDto` which [contains `OutputMetadata`](https://github.com/iotaledger/iota.rs/blob/595da10d75f3d746a45c8dfda6561399ab71a435/client/src/secret/types.rs#L159). Note that the latter is not the `Dto` type. That results in additional fields being present in the resulting JSON and some required ones missing. I assume this should be changed to `OutputMetadataDto`?
2. `InputSigningDataDto` serializes its [`output_metadata` field as `outputMetadata`](https://github.com/iotaledger/iota.rs/blob/595da10d75f3d746a45c8dfda6561399ab71a435/client/src/secret/types.rs#L158) but in NodeJS this is defined as [`outputMetaData`](https://github.com/iotaledger/iota.rs/blob/595da10d75f3d746a45c8dfda6561399ab71a435/client/bindings/nodejs/types/preparedTransactionData.ts#L40) (note the different capitalization).
3. `InputSigningDataDto` contains an [`Option<Chain>` field](https://github.com/iotaledger/iota.rs/blob/595da10d75f3d746a45c8dfda6561399ab71a435/client/src/secret/types.rs#L159) which is a newtype wrapper around `Vec<Segment>`. That serialization is not entirely compatible with the corresponding [TS definition](https://github.com/iotaledger/iota.rs/blob/595da10d75f3d746a45c8dfda6561399ab71a435/client/bindings/nodejs/types/preparedTransactionData.ts). Note `ISegment`'s `bs` field which is a `Uint8Array`. The issue is that `Uint8Array`'s serialization is different from Rust arrays. The latter expects JSON like `[1,2,3,4]` while `Uint8Array` produces `{"0":1,"1":2,"2":3,"3":4}`. This leads to deserialization problems when sending JSON over the message interface and attempting to deserialize it into a `Message`. A potential fix might be to use `number[]`. Trying to deserialize anything outside the byte range into `[u8; 4]` would be caught by serde, but double checking that would be good.

### Can the issue reliably be reproduced?

Yes

### Steps to reproduce the issue

1. Run steps 1 and 2 of the offline signing example.
2. Copy the contents of `prepared_transaction.json` into an example file in the NodeJS bindings and try assigning it to some constant of type `IPreparedTransactionData`.
3. Notice the errors. To see the above-mentioned errors, one has to deal with the `bs` fields not being of type `Uint8Array` first. That can be temporarily fixed by wrapping these in `Uint8Array.from` (or deleting them).

### Errors

_No response_

### Duplicate declaration

- [X] I have searched the issues tracker this issue and there is none

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Thoralf-M:
# Description of change

Use OutputMetadataDto instead of OutputMetadata

Fixes https://github.com/iotaledger/iota.rs/issues/1547

## Type of change

Running offline signing examples

Philipp wrote `Trying to deserialize anything outside the byte range into [u8; 4] would be caught by serde, but double checking that would be good.` in the issue, so I tried to run it with `"bs": [128, 0, 0, 44, 1]` instead of `"bs": [128, 0, 0, 44]` and it returned `Error("trailing characters", line: 1, column: 1644)`, so should be good

## Change checklist

### DIFF
--- a/client/bindings/nodejs/examples/offline_signing/1_transaction_preparation.ts
+++ b/client/bindings/nodejs/examples/offline_signing/1_transaction_preparation.ts
@@ -1,7 +1,7 @@
 // Copyright 2021-2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-import { Client, initLogger, IPreparedTransactionData } from '@iota/client';
+import { Client, initLogger } from '@iota/client';
 import { writeFile, readFile } from 'fs/promises';
 
 require('dotenv').config({ path: '../../../../.env' });
--- a/client/bindings/nodejs/examples/offline_signing/1_transaction_preparation.ts
+++ b/client/bindings/nodejs/examples/offline_signing/1_transaction_preparation.ts
@@ -1,10 +1,10 @@
 // Copyright 2021-2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-import { Client, initLogger } from '@iota/client';
+import { Client, initLogger, IPreparedTransactionData } from '@iota/client';
 import { writeFile, readFile } from 'fs/promises';
 
-require('dotenv').config({ path: '../../../.env' });
+require('dotenv').config({ path: '../../../../.env' });
 
 // From examples directory, run with:
 // node ./dist/offline_signing/1_transaction_preparation.js
--- a/client/bindings/nodejs/types/preparedTransactionData.ts
+++ b/client/bindings/nodejs/types/preparedTransactionData.ts
@@ -37,7 +37,7 @@ export interface IInputSigningData {
     /**
      * The output metadata
      */
-    outputMetaData: IOutputMetadataResponse;
+    outputMetadata: IOutputMetadataResponse;
     /**
      * The chain derived from seed, only for ed25519 addresses
      */
@@ -65,5 +65,5 @@ export interface IRemainder {
 }
 export interface ISegment {
     hardened: boolean;
-    bs: Uint8Array;
+    bs: number[];
 }
--- a/client/src/secret/types.rs
+++ b/client/src/secret/types.rs
@@ -6,7 +6,10 @@
 use crypto::keys::slip10::Chain;
 use iota_types::block::{
     address::Address,
-    output::{dto::OutputDto, Output, OutputId, OutputMetadata},
+    output::{
+        dto::{OutputDto, OutputMetadataDto},
+        Output, OutputId, OutputMetadata,
+    },
 };
 use serde::{Deserialize, Serialize};
 #[cfg(feature = "stronghold")]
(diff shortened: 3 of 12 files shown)


client/bindings/nodejs/examples/offline_signing/1_transaction_preparation.ts
```diff
@@ -1,10 +1,10 @@
 // Copyright 2021-2022 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-import { Client, initLogger } from '@iota/client';
+import { Client, initLogger, IPreparedTransactionData } from '@iota/client';
```
thibault-martinez: Warning

client/src/secret/types.rs
```diff
@@ -156,7 +159,7 @@ pub struct InputSigningDataDto {
     pub output: OutputDto,
     /// The output metadata
     #[serde(rename = "outputMetadata")]
-    pub output_metadata: OutputMetadata,
+    pub output_metadata: OutputMetadataDto,
```
thibault-martinez, line 162: Are we 100% sure this doesn't cause any issue with stored data ?

client/bindings/nodejs/types/preparedTransactionData.ts
```diff
@@ -65,5 +65,5 @@ export interface IRemainder {
 }
 export interface ISegment {
     hardened: boolean;
-    bs: Uint8Array;
+    bs: number[];
```
thibault-martinez, line 68: Maybe elaborate the changelog a tiny bit more

client/src/secret/types.rs
Thoralf-M: Checked in the wallet, there we only use `InputSigningData` and not the DTO and also that we only create during the input selection, but don't store it

client/bindings/nodejs/types/preparedTransactionData.ts
Thoralf-M: Please check now

thibault-martinez (approved):

kwek20 (approved):

PR #1570, opened 2023-02-10
ISA improvements

thibault-martinez:
## Change checklist

Tick the boxes that are relevant to your changes, and delete any items that are not.

- [ ] I have followed the contribution guidelines for this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that new and existing unit tests pass locally with my changes

PR #1571, opened 2023-02-10
Signing tests

Thoralf-M:
Will rebase after https://github.com/iotaledger/iota.rs/pull/1570 is merged


client/tests/signing/basic.rs
```diff
@@ -0,0 +1,349 @@
...
+        remainder: None,
+    };
+
+    let unlocks = secret_manager
+        .sign_transaction_essence(&prepared_transaction_data)
+        .await?;
+
+    assert_eq!(unlocks.len(), 3);
+    assert_eq!((*unlocks).get(0).unwrap().kind(), SignatureUnlock::KIND);
+    assert_eq!((*unlocks).get(1).unwrap().kind(), ReferenceUnlock::KIND);
```
thibault-martinez: Maybe do `matches!` on the unlock to also assert the index ?

client/tests/signing/mod.rs
```diff
@@ -0,0 +1,5 @@
+// Copyright 2023 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+mod alias;
```
thibault-martinez, line 4: Any reason not to add nft in this PR ?

client/tests/signing/alias.rs
```diff
@@ -0,0 +1,339 @@
...
+        remainder: None,
+    };
+
+    let unlocks = secret_manager
+        .sign_transaction_essence(&prepared_transaction_data)
+        .await?;
+
+    assert_eq!(unlocks.len(), 2);
+    assert_eq!((*unlocks).get(0).unwrap().kind(), SignatureUnlock::KIND);
+    assert_eq!((*unlocks).get(1).unwrap().kind(), AliasUnlock::KIND);
```
thibault-martinez: matches! ?

client/tests/signing/mod.rs
Thoralf-M: Would like to have it merged rather sooner because the changes to the others tests can be conflicting and I also want to add quite a few more tests which would make the PR way bigger

client/tests/signing/basic.rs
Thoralf-M: Can you show an example? Not sure what you mean, isn't `.get(1).unwrap()` basically doing that?

client/tests/signing/mod.rs
thibault-martinez: Fair enough

client/tests/signing/basic.rs
thibault-martinez: I meant something like `assert!(matches!(*unlocks.get(1).unwrap(), ReferenceUnlock(0)));`. But that's actually not possible because the field is not pub. But atm you are not checking the inner index. Maybe add another assert for it ?

thibault-martinez (approved):

kwek20 (approved):

PR #1574, opened 2023-02-13
Cleanup setup.py

Thoralf-M:


client/bindings/python/setup.py
```diff
@@ -22,10 +22,9 @@ def get_py_version_cfgs():
 
 setup(
     name="iota_client",
-    version="1.0.0.rc.2",
+    version="1.0.0-rc.2",
```
thibault-martinez, line 25: It's not been released yet, right ?

Thoralf-M: It is, next will be rc.3

thibault-martinez (approved):

PR #1584, opened 2023-02-16
Add more fulfil amount combinations to ISA

thibault-martinez:
## Change checklist

### DIFF
--- a/client/tests/input_selection/alias_outputs.rs
+++ b/client/tests/input_selection/alias_outputs.rs
@@ -1411,26 +1411,22 @@ fn two_aliases_required() {
     assert!(unsorted_eq(&selected.inputs, &inputs));
     assert_eq!(selected.outputs.len(), 3);
     assert!(selected.outputs.contains(&outputs[0]));
-    assert!(
-        selected
-            .outputs
-            .iter()
-            .any(|output| if let Output::Alias(output) = output {
-                output.alias_id() == &alias_id_1
-            } else {
-                false
-            })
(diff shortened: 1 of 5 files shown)


client/tests/input_selection/alias_outputs.rs
```diff
@@ -1993,3 +1996,47 @@ fn remainder_address_in_governor() {
...
+    )
+    .select();
+
+    assert!(matches!(
+        selected,
+        Err(Error::InsufficientAmount {
+            found: 2_000_000,
+            required: 3_000_000,
+        })
+    ));
```
Thoralf-M, line 2041: I think that's wrong or at least confusing
Maybe we should store the amount of provided outputs and also return that, because only a basic output with 1_000_000 was provided, why is now 3_000_000 required? 3_000_000 makes sense if you need to transition the alias, but if we can't use any amount, why is it even added to the `found`?

thibault-martinez: Yeah but the alias is actually required because of the basic sender

thibault-martinez: That's the only way to get an automatic governance transition

Thoralf-M: Ah missed the sender thing

Thoralf-M (dismissed):

kwek20 (approved):

Thoralf-M (approved):

PR #1587, opened 2023-02-17
ISA: Avoid second run of amount requirement

thibault-martinez:
## Change checklist

### DIFF
--- a/client/src/api/block_builder/input_selection/core/requirement/amount.rs
+++ b/client/src/api/block_builder/input_selection/core/requirement/amount.rs
@@ -3,7 +3,7 @@
 
 use std::collections::HashMap;
 
-use super::{InputSelection, Requirement};
+use super::InputSelection;
 use crate::{
     block::{
         address::Address,
@@ -11,6 +11,7 @@ use crate::{
             unlock_condition::StorageDepositReturnUnlockCondition, AliasOutputBuilder, AliasTransition,
             FoundryOutputBuilder, NftOutputBuilder, Output, OutputId, Rent,
         },
+        protocol::ProtocolParameters,
     },
     error::{Error, Result},
     secret::types::InputSigningData,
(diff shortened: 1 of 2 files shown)


client/src/api/block_builder/input_selection/core/requirement/amount.rs
```diff
@@ -147,6 +150,14 @@ impl AmountSelection {
                 *self.inputs_sdr.entry(*sdruc.return_address()).or_default() += sdruc.amount();
             }
 
+            if !input.output.is_basic() {
```
Thoralf-M, line 153: What about aliases with governance transition? Then we can't do this, no?

Thoralf-M, line 153: Just to confirm: alias/foundries/nfts we want to burn are selected already earlier and then can't be added here anymore, is that true?

thibault-martinez: This is only evaluated on available inputs so if a governance transition was already provided in output or required because of a sender feature, by the time of amount selection it wouldn't be available anymore. And for amount we give a None (L162) as transition which will default to State

client/src/api/block_builder/input_selection/core/requirement/amount.rs
```diff
@@ -331,37 +346,13 @@ impl InputSelection {
...
-                // TODO explanation of Amount
-                self.requirements.push(Requirement::Amount);
-
-                return Ok(amount_selection.into_newly_selected_inputs());
+            if self.fulfil(inputs, &mut amount_selection) {
+                break 'fulfil;
             }
         }
 
-        if amount_selection.missing_amount() != 0 {
```
kwek20, line 361: should we make this an assert or something? Just in case theres a mistake

thibault-martinez: It's just been moved, it's still checked

PR #1612, opened 2023-02-27
Java tests CI

### LINKED ISSUE #1588: Fix Java bindings CI

Thoralf-M:
# Description of change

Change requestTestFundsFromFaucet to reduce the faucet requests and run test with the private tangle

Fixes #1588 

## Type of change

Describe the tests that you ran to verify your changes.

Make sure to provide instructions for the maintainer as well as any relevant configurations.

## Change checklist

### DIFF
--- a/.github/workflows/examine-java-bindings.yml
+++ b/.github/workflows/examine-java-bindings.yml
@@ -75,33 +75,27 @@ jobs:
       - name: Validate Gradle wrapper
         uses: gradle/wrapper-validation-action@e6e38bacfdf1a337459f332974bb2327a31aaf4b
 
-      - name: Install docker (macos)
-        if: ${{ startsWith(matrix.os, 'macos') }}
-        run: |
-          brew install docker
-          colima start
-
       - name: Start private tangle
         uses: './.github/actions/private-tangle/setup'
-        if: ${{ !startsWith(matrix.os, 'windows') }}
+        if: ${{ startsWith(matrix.os, 'ubuntu') }}
 
-      - name: Run Gradle Test Windows
-        if: ${{ startsWith(matrix.os, 'windows') }}
+      - name: Run Gradle Test (macos, windows)
+        if: ${{ !startsWith(matrix.os, 'ubuntu') }}
         working-directory: client/bindings/java/
         run: |
           chmod +x gradlew
-          # Only run test that doesn't require a node because we can't use the private tangle on windows
+          # Only run test that doesn't require a node because we can't use the private tangle script on macos and windows
           ./gradlew test -i --tests "*testAddressDerivation"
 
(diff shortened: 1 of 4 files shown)


.github/workflows/examine-java-bindings.yml
```diff
@@ -78,7 +78,8 @@ jobs:
         working-directory: client/bindings/java/
         run: |
           chmod +x gradlew
-          ./gradlew test -i
+          # Only run test that doesn't require a node 
+          ./gradlew test -i --tests "*testAddressDerivation"
```
thibault-martinez, line 88: But we're just ignoring tests now? That doesn't sound like solving the actual issue :(

Thoralf-M: The main issue was that we sent too many requests to the faucet (the rate limit got changed, probably that's why we get rate limited more easily now)
Already reduced the amount of requests in requestTestFundsFromFaucet, but in the logs you can still see
```
    null
    {"error":{"code":"500","message":"internal server error. error: %!s(\u003cnil\u003e)"}}
```
for the faucet requests which is returned when we're rate limited, of course I could increase maxAttemts to 1000 or something so it will work at some point, but not sure if that's really better then

Also the Rust tests should already cover almost everything, so I think we don't need to test the Java tests in the CI every time

thibault-martinez: And we're 100% sure we're not spamming somehow? If the rust tests work, the Java test should work as well...

thibault-martinez: At the very least I don't think we should do `--tests "*testAddressDerivation"`, it seems very error prone if we add more tests. Can't we just ignore some tests?

Thoralf-M: I think we can add `@ignore`, but we would need to add this to basically all tests and I haven't seen a way to run them then locally

thibault-martinez: I understand it's annoying but it really feels like the way to go tbh :/
I prefer having to put this everywhere than not running actual tests that potentially fail

Thoralf-M: But if we can't even run it (I'm not 100% sure, but haven't seen any option to do so), then we could also completely remove them since they would never be executed unless you comment this

Thoralf-M: Changed it so we run all tests with the private tangle on ubuntu and only the address test on macos and windows because the private tangle setup doesn't easily work there
This way all things should be tested on ubuntu, but we still know that it also builds on macos and windows, so I think that's okay

thibault-martinez (approved):

PR #1632, opened 2023-03-06
Update python client constructor to be more Pythonic

### LINKED ISSUE #48: Create pythonic constructor

Dr-Electron:
# Description of change

Updated the python client constructor to be more Pythonic. Also added a docstring. Ideally if we decide to use the jsonschema tool from iotaledger/iota-sdk#41 maybe some of the descriptions and validations from there can be used 🤔 

Right now it would also be possible to use auth for Nodes (it would look something like this:)
```
client = IotaClient(
    nodes=[
        'https://api.testnet.shimmer.network',
        {"url":"https://api.testnet.shimmer.network","auth":{"jwt":"token","basicAuthNamePwd":["user","name"]},"disabled":True}
#        {'url':'https://api.testnet.shimmer.network'}
    ])
```
 but I didn't decide on a format yet and maybe will do that in another PR 😅 

TODOs:

- [x] Discuss __default() solution for using defaults where the Rust`Option` could possibly be different than the Rust `None` (and therefore we can't use the Python None to check)
- [x] Check if docs need changes

Links to any relevant issues

Fixes iotaledger/iota-sdk#48 

## Type of change

The generated json was validated with the jsonschema generated by my tool

## Change checklist


client/bindings/python/iota_client/client.py
```diff
@@ -4,13 +4,96 @@
...
 from json import dumps
+import humps
+from datetime import timedelta
 
 class IotaClient(NodeCoreAPI, NodeIndexerAPI, HighLevelAPI, Utils):
-    def __init__(self, client_config=None):
+    __default = object()
+    def __init__(
+        self,
+        node = __default,
```
Thoralf-M: Why have this node?

Dr-Electron: I actually thought about it. I started "mirroring" the Rust builder system to see which parameters "can" be set in Rust.

And therefore I added `node`. Another argument for it, is that most people are probably used to a syntax like:
```python
client = IotaClient('url')
#or
client = IotaClient(node='url')
```

instead of

```python
client = IotaClient(['url'])
#or
client = IotaClient(nodes=['url'])
```

But on the other hand maybe we should promote using the quorum feature more anyway (as this should be best practice (in crypto) anyway),

Or just make a single parameter that accepts both 🤔 

So I will think about changing it 👍

Dr-Electron: Reimplemented this part. Now `nodes` can be: `string | array of strings | Node | array of Node`

Dr-Electron:
It's now also possible to add Nodes with auth, by doing something like:

```python
from iota_client import IotaClient
from iota_client import Node

 # Create an IotaClient instance
 client = IotaClient('https://api.testnet.shimmer.network', Node('https://api.testnet.shimmer.network', 'jwt_token', 'username', 'password', disabled=True))
```

Thoralf-M (changes requested):
Missing changelog entry

client/bindings/python/iota_client/client.py
```diff
@@ -4,13 +4,102 @@
...
+            A single Node URL or an array of URLs.
+        primary_node : string
+            Node which will be tried first for all requests.
+        primary_pow_node : string
+            Node which will be tried first when using remote PoW, even before the primary_node.
+        permanode : string
+            Permanode URL.
+        ignore_node_health : bool
+            If the node health should be ignored.
+        nodes : array of strings
```
Thoralf-M: Duplicate of L35

client/bindings/python/iota_client/client.py
```diff
@@ -190,3 +279,44 @@ def submit_payload(self, payload_dto):
...
+        """Initialize a Node
+
+        Parameters
+        ----------
+        url : string
+            Node url
+        jwt : string
+            JWT token
+        user : string
+            User for basic authentication
```
Thoralf-M:
```suggestion
        username : string
            Username for basic authentication
```

Thoralf-M: Would need to be changed also below

Dr-Electron:
> Missing changelog entry

So I would just add something like this, right?
```
## Unreleased - 2023-MM-DD

### Added

- `Node` class

### Changed

- `IotaClient` now uses multiple arguments instead of a single dictionary
```

client/bindings/python/iota_client/client.py
Dr-Electron: Good catch. Forgot to remove it 👍

client/bindings/python/iota_client/client.py
Dr-Electron: Done

Thoralf-M (dismissed):

client/bindings/python/iota_client/client.py
```diff
@@ -4,13 +4,99 @@
...
+        pow_worker_count : int
+            The amount of threads to be used for proof of work.
         """
-        if client_config:
-            client_config = dumps(client_config)
+        client_config = dict(locals())
+        del client_config['self']
+
+        if isinstance(nodes, list):
+            _nodes = nodes
```
Thoralf-M: Can't you just iter over `nodes` directly?

Dr-Electron: Wdym? nodes can be a string or a list of strings 🤔

Thoralf-M: Why do you do `_nodes = nodes` instead of something like this?
```python
for i, node in enumerate(nodes):
    if isinstance(node, Node):
            nodes[i] = node.as_dict()
        else:
            nodes[i] = node
```

Dr-Electron: I'm not a python dev 😆
What you suggest should work. Will change it

Dr-Electron: Hmm, wait. This won't work.

I you call the client like: `client = IotaClient('node-url')` you pass an array of char.
That's why I checked if it is a list first 🤔

Thoralf-M: You can still check it, but I don't see for what you need to create this new `_nodes` variable `_nodes = nodes`

Dr-Electron: Tbh currently I don't have any idea why I did it 😅 .
Will check again later

Dr-Electron: Done

Dr-Electron:
Closed in favour of https://github.com/iotaledger/iota-sdk/pull/25

PR #1633, opened 2023-03-06
CI improvements

thibault-martinez:


.github/workflows/bindings-java.yml
```diff
@@ -1,34 +1,28 @@
...
     paths:
-      - ".github/workflows/examine-java-bindings.yml"
+      - ".github/workflows/bindings-java.yml"
       - ".github/actions/**"
       - "client/bindings/java/**"
-      # Watch the Rust core too
-      - "client/src/**"
-      - "client/Cargo.toml"
-      - "Cargo.lock"
+      - "**.rs"
```
Thoralf-M: With this is would also run if we change something in the Rust part of nodejs, no?

thibault-martinez: Ah yeah right. Do you think it's better to list all relevant rust file paths or to have `**.rs` and exclude what is not relevant? I fear that with listing all relevant, we end up forgetting to add new code. Like we forgot to add `types`.

Thoralf-M: Exclude sounds like the safer way

.github/workflows/bindings-java.yml
```diff
@@ -54,7 +48,7 @@ jobs:
           cache: true
           cache-root: client/bindings/java/lib/native/
           cache-job-id: ${{ github.workflow }}-${{ github.job }}-${{ matrix.os }}-${{ matrix.java }}
```
Thoralf-M, line 56: Does it make sense to remove the language specific caches now? `-${{ matrix.java }}`

thibault-martinez: Yeah maybe

thibault-martinez: But if you allow me I'll do a dedicated PR for caching, there are too many things to explore and I need to chat with Rajiv.

.github/workflows/bindings-java.yml
thibault-martinez: Done

Thoralf-M (approved):

kwek20 (approved):

PR #1642, opened 2023-03-09
Unify "Generate Addresses" examples

### LINKED ISSUE #1603: Unify "Generate Addresses" examples
## Description

The [Generate Addresses](https://wiki.iota.org/shimmer/iota.rs/how_tos/generate_addresses/) examples explain different features in different languages.
This issue is part of the efforts in iotaledger/iota-sdk#59 

## Motivation

We should show/explain the same features in all languages (if they exist).

## Requirements

Let's show in every example how to:

- [x] Generate Addresses with default account index and range
    - [x] Rust
    - [x] Node.js
    - [x] Python
    - [x] Java
- [x] Generate Addresses with custom account index and range
    - [x] Rust
    - [x] Node.js
    - [x] Python
    - [x] Java (It exists, but just to be complete we could also show how to define the custom account index)
- [x] Generate Addresses offline
    - [x] Rust
    - [x] Node.js
    - [x] Python
    - [x] Java
- [x] Generate internal Address
    - [x] Rust
    - [x] Node.js
    - [x] Python
    - [x] Java

## Open questions (optional)

If something has to be implemented for this to work that would take too much time, would it make sense to open an issue for it? That way we can maybe call for action in the community if we know which features exactly are missing in the bindings.

Yes.

Dr-Electron:
# Description of change

Unified the examples by making sure the all explain the same features and have a similar output

fixes #1603 

## Type of change

- Documentation Fix

Run all examples

## Change checklist

- [x] I have made corresponding changes to the documentation


Thoralf-M (changes requested):
fmt failed

client/bindings/java/examples/src/GenerateAddresses.java
```diff
@@ -11,13 +11,54 @@ public static void main(String[] args) throws ClientException, InitializeClientE
...
+            .withBech32Hrp("rms"));
+
+        // Print the addresses.
+        System.out.println("List of offline generated public addresses:");
+        for (String address : offlineGeneratedAddresses) {
+            System.out.println(address);
+        }
+        System.out.println();
     }
 }
```
Dr-Electron:
```suggestion
}

```

client/bindings/python/examples/02_generate_addresses.py
```diff
@@ -5,17 +5,52 @@
...
+    "accountIndex": 0,
+    "range": {
+        "start": 0,
+        "end": 4,
+    },
+    "internal": True,
+})
+
+print('List of generated internal addresses:', *addresses, sep='\n')
+print()
```
Thoralf-M, line 39: Do you really think that this is useful in this example? I think 99% don't care about these and would be fine with just the public addresses and if someone really cares for whatever reason, he should be able to find this option himself

Dr-Electron: I think every example is useful. We should split up examples later on and sort them. But I personally am a big fan if everything is part of examples/how-tos. That's why they exist. And basically no downside

Thoralf-M: > I think every example is useful.

Sure many examples are great, but I think there should be some smaller examples without too much going on for people to get started and then we can have many more examples with more things (they shouldn't be in the same how tos section, otherwise it's also too many examples so you don't even know where to start)
> We should split up examples later on 

Why later and not now already?

Dr-Electron: Because we don't really have the tool yet. So we will need to wait for the next sprint

Thoralf-M (approved):

thibault-martinez (approved):

